PR #16, opened 2018-07-03
Class 1 dummy

yaronweiss:
Pull 1


yaronweiss:
it was great test and work as expected

PR #45, opened 2018-08-06
Find And Replace fixed issues

yulev55:


Ginger/Ginger/Actions/ActionEditPage.xaml.cs
```diff
@@ -800,6 +800,17 @@ public bool ShowAsWindow(eWindowShowStyle windowStyle = eWindowShowStyle.Free, b
...
+                    Button saveButton = new Button();
+                    saveButton.Content = "Save";
+                    saveButton.Click += new RoutedEventHandler(saveButton_Clicked);
+                    Button uncoButton = new Button();
+                    uncoButton.Content = "Undo & Close";
+                    uncoButton.Click += new RoutedEventHandler(undoBtn_Click);
+                    winButtons.Add(saveButton);
+                    winButtons.Add(uncoButton);
+                    break;
                 case General.RepositoryItemPageViewMode.View:
```
MeniKadosh1: why "Standalone" was needed to be added? for which scenario it is needed? it supposed to be Standalone only when exsit in Shared Repository and you already have sace for it.

Ginger/Ginger/Activities/ActivityEditPage.xaml.cs
```diff
@@ -177,6 +177,17 @@ public bool ShowAsWindow(eWindowShowStyle windowStyle = eWindowShowStyle.Dialog,
...
+                    Button saveButton = new Button();
+                    saveButton.Content = "Save";
+                    saveButton.Click += new RoutedEventHandler(saveBtn_Click);
+                    Button undoButton = new Button();
+                    undoButton.Content = "Undo & Close";
+                    undoButton.Click += new RoutedEventHandler(undoBtn_Click);
+                    winButtons.Add(undoButton);
+                    winButtons.Add(saveButton);
+                    break;
                 case General.RepositoryItemPageViewMode.View:
```
MeniKadosh1: why "Standalone" was needed to be added? for which scenario it is needed? it supposed to be Standalone only when exsit in Shared Repository and you already have sace for it.

Ginger/Ginger/Functionalities/FindAndReplacePage.xaml.cs
```diff
@@ -756,9 +845,11 @@ public void ViewAction(FoundItem actionToView)
...
             if(mContext == eContext.RunsetPage)
                 w = new ActionEditPage(act, General.RepositoryItemPageViewMode.View);
+            else if(mContext == eContext.AutomatePage)
+                w = new ActionEditPage(act, General.RepositoryItemPageViewMode.Automation);
             else if (Parent is BusinessFlow)
-                w = new ActionEditPage(act, General.RepositoryItemPageViewMode.Child, Parent as BusinessFlow);
-            else if(Parent is Activity)
+                w = new ActionEditPage(act, General.RepositoryItemPageViewMode.Standalone, Parent as BusinessFlow);
+            else if (Parent is Activity)
                 w = new ActionEditPage(act, General.RepositoryItemPageViewMode.Child, actParentActivity: Parent as Activity);// add save handeling to save activity if not null
```
MeniKadosh1: I think no matter if the Parent to save it BF or Activity it supposed to be loaded in "Child" mode no?

Ginger/Ginger/Actions/ActionEditPage.xaml.cs
yulev55: The current case for .child show a "save" button as we need,
But show a message regarding the analyzer, "Saving auto fixed issues..." which is not correct in our case.

So yes we need the same functionality as for the .child case - only save button
but the message when asking if to save is not correct, so I have added another case for it - standalone. maybe the name "standalone" is not correct, or I can check something else before showing the message if we need to show message for the analyzer or just regular save message:

        private void saveBtnAnalyzer_Click(object sender, RoutedEventArgs e)
        {
            if (Reporter.ToUser(eUserMsgKeys.SaveAnalyzerItemWarning) == MessageBoxResult.Yes)
            {
                Reporter.ToGingerHelper(eGingerHelperMsgKey.AnalyzerSavingFixedIssues, null, mAction.Description);
                saveWasDone = true;
                mActParentBusinessFlow.Save();
                Reporter.CloseGingerHelper();
            }

            IsPageClosing = true;
            _pageGenericWin.Close();
        }

Ginger/Ginger/Activities/ActivityEditPage.xaml.cs
yulev55: the same as above,
also the shared repository is not correct because of the message shown and the saving process.

Ginger/Ginger/Functionalities/FindAndReplacePage.xaml.cs
yulev55: the same as two above.
needed to ad another case

Ginger/Ginger/Actions/ActionEditPage.xaml.cs
```diff
@@ -904,14 +881,16 @@ private void saveBtn_Click(object sender, RoutedEventArgs e)
...
         {
-            if (Reporter.ToUser(eUserMsgKeys.SaveAnalyzerItemWarning) == MessageBoxResult.Yes)
+            if ((mActParentBusinessFlow != null && Reporter.ToUser(eUserMsgKeys.SaveItemParentWarning, GingerDicser.GetTermResValue(eTermResKey.BusinessFlow),mActParentBusinessFlow.Name) == MessageBoxResult.Yes) 
+                || (mActParentActivity != null && Reporter.ToUser(eUserMsgKeys.SaveItemParentWarning, GingerDicser.GetTermResValue(eTermResKey.Activity), mActParentActivity.ActivityName) == MessageBoxResult.Yes))
             {
-                Reporter.ToGingerHelper(eGingerHelperMsgKey.AnalyzerSavingFixedIssues, null, mAction.Description);
                 saveWasDone = true;
-                mActParentBusinessFlow.Save();
-                Reporter.CloseGingerHelper();
+                if(mActParentBusinessFlow != null)
```
MeniKadosh1: just please move "saveWasDone = true;" to after save

Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -472,7 +472,7 @@ private async void SaveAllFixedItems_Click(object sender, RoutedEventArgs e)
             }
             // TODO: click/use the same code on solution which will save all changed items...
             // Meanwhile the below is good start 
-            if (Reporter.ToUser(eUserMsgKeys.SaveAnalyzerItemWarning) == MessageBoxResult.Yes)
+            if (Reporter.ToUser(eUserMsgKeys.SaveItemParentWarning) == MessageBoxResult.Yes)
             {
```
MeniKadosh1: need to set the meessage params here

Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -472,7 +472,7 @@ private async void SaveAllFixedItems_Click(object sender, RoutedEventArgs e)
             }
             // TODO: click/use the same code on solution which will save all changed items...
             // Meanwhile the below is good start 
-            if (Reporter.ToUser(eUserMsgKeys.SaveItemParentWarning) == MessageBoxResult.Yes)
+            if (Reporter.ToUser(eUserMsgKeys.SaveItemParentWarning, GingerDicser.GetTermResValue(eTermResKey.BusinessFlow)) == MessageBoxResult.Yes)
```
MeniKadosh1: need to point to the new message added

Ginger/GingerCore/Reporter/UserMessagesPool.cs
```diff
@@ -198,7 +198,8 @@ public static void LoadUserMessgaesPool()
...
             Reporter.UserMessagesPool.Add(eUserMsgKeys.AnalyzerSaveRunSet, new UserMessage(eMessageType.WARN, "Issues Detected By Analyzer", "Please save the " + GingerDicser.GetTermResValue(eTermResKey.RunSet) + " first", MessageBoxButton.OK, MessageBoxResult.None));
             Reporter.UserMessagesPool.Add(eUserMsgKeys.SaveItemParentWarning, new UserMessage(eMessageType.WARN, "Item Parent Save", "Save process will actually save the item {0} parent which called '{1}'." + System.Environment.NewLine + System.Environment.NewLine + "To continue with save?", MessageBoxButton.YesNo, MessageBoxResult.No));
-            
+            Reporter.UserMessagesPool.Add(eUserMsgKeys.SaveAllItemsParentWarning, new UserMessage(eMessageType.WARN, "Items Parent Save", "Save process will actually save all items {0} parent." + System.Environment.NewLine + System.Environment.NewLine + "To continue with save?", MessageBoxButton.YesNo, MessageBoxResult.No));
+
```
MeniKadosh1: need to change the message as disscused

Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -472,7 +472,7 @@ private async void SaveAllFixedItems_Click(object sender, RoutedEventArgs e)
             }
             // TODO: click/use the same code on solution which will save all changed items...
             // Meanwhile the below is good start 
-            if (Reporter.ToUser(eUserMsgKeys.SaveItemParentWarning, GingerDicser.GetTermResValue(eTermResKey.BusinessFlow)) == MessageBoxResult.Yes)
+            if (Reporter.ToUser(eUserMsgKeys.SaveAllItemsParentWarning, GingerDicser.GetTermResValue(eTermResKey.BusinessFlow)) == MessageBoxResult.Yes)
             {
```
MeniKadosh1: now need to remove the un needed mesage param sent

PR #49, opened 2018-08-07
Implemented the code to fix the Bug_5359

shahanemahesh:


MeniKadosh1:
please do not push before beta process ends.

shahanemahesh:
User have not created his own function, he have used  the vbs function already in our vbs function "Check if date valid", so in this case then will have to change our vbs function, please let me know so we need to do that?

Sent from Nine
________________________________

JaspreetGIT:
Let's update the VB Expression which is getting added while selecting IsDate from VB Functions

shahanemahesh:
+Jaspreet

Sent from Nine
________________________________

shahanemahesh:
As discussed this needs to be handled in different way so deleting the branch

PR #55, opened 2018-08-10
Feature/run plsql file from ginger

JaspreetGIT:


MeniKadosh1:
Jinendra please review and see if we should push it our self because Jaspreet is OOO

JinendraGhodke:
Merged and Deleting branch

PR #56, opened 2018-08-10
Adding remove special chars

JaspreetGIT:


JaspreetGIT:
I added only those characters which are special either for our value expression or for VBS functions. For $ and % there are issues as such. So not added

MeniKadosh1:
Jinendra please see if we should push it our self because Jaspreet is OOO

JinendraGhodke:
Merged and deleting a branch. Please delete from your local.

PR #57, opened 2018-08-10
Modified the code for new Dirtystatus Implemention on DataSource

shahanemahesh:


shahanemahesh:
this changes are for Bug:6219-- #4 point

JaspreetGIT:
We also need the dirty status update for data table to know which tables are modified and to commit during commit all. you can push this for now. I will handle DataTable part

PR #65, opened 2018-08-13
Fixing bug 6291

OlegOrlovsky:


Ginger/Ginger/Actions/AddActionPage.xaml.cs
```diff
@@ -110,7 +110,9 @@ private void LoadGridData()
                     {
                         LegacyActions.Add(cA);
                     }
```
MeniKadosh1: why we still have the IsLegacyAction flag? we said we will replace it with the LegacyActionPlatformsList but currently we have both. please fully replace IsLegacyAction  with LegacyActionPlatformsList

MeniKadosh1:
Oleg, i merged it and deleted the branch, please delete it also from your local

PR #86, opened 2018-08-22
Cassandra DB connection fix

KashmiraDeshpande:


Ginger/GingerCore/Environments/Database.cs
```diff
@@ -394,7 +394,10 @@ public Boolean Connect(bool displayErrorPopup = false)
 
                     case eDBTypes.Cassandra:
                         GingerCassandra CassandraDriver = new GingerCassandra(this);
-                        CassandraDriver.Connect();
+                        bool cassandraConnection;
+                        cassandraConnection= CassandraDriver.Connect();
+                        if (cassandraConnection == true)
```
JinendraGhodke: If connected successfully then you need to update last connection used time
LastConnectionUsedTime = DateTime.Now;

Ginger/GingerCore/Environments/Database.cs
```diff
@@ -394,7 +394,10 @@ public Boolean Connect(bool displayErrorPopup = false)
 
                     case eDBTypes.Cassandra:
                         GingerCassandra CassandraDriver = new GingerCassandra(this);
-                        CassandraDriver.Connect();
+                        bool cassandraConnection;
```
JinendraGhodke: cassandraConnection is more relevant if it is connection object.
As it is bool please rename to something like isConnected

PR #88, opened 2018-08-23
Feature/6090 need solution to have linking between existing tc in tdp & existing ginger scripts

uzanh:


Ginger/Ginger/SolutionWindows/TreeViewItems/BusinessFlowsFolderTreeItem.cs
```diff
@@ -365,7 +365,7 @@ void ITreeViewItem.SetTools(ITreeView TV)
                 MenuItem importMenu = TreeViewUtils.CreateSubMenu(mContextMenu, "Import");
                 TreeViewUtils.AddSubMenuItem(importMenu, "Import External " + GingerDicser.GetTermResValue(eTermResKey.BusinessFlow), ImportExternalBuinessFlow, null, "@Import_16x16.png");
                 TreeViewUtils.AddSubMenuItem(importMenu, "Import ALM Test Set", ALMTSImport, null, "@ALM_16x16.png");
-                TreeViewUtils.AddSubMenuItem(importMenu, "Import ALM Test Set By ID", ALMTSImportById, null, "@ALM_16x16.png");
+                if(App.UserProfile.Solution.AlmType != ALMIntegration.eALMType.QC) TreeViewUtils.AddSubMenuItem(importMenu, "Import ALM Test Set By ID", ALMTSImportById, null, "@ALM_16x16.png");
                 TreeViewUtils.AddSubMenuItem(importMenu, "Import Gherkin Feature File", ImportGherkinFeature, null, "@FeatureFile_16X16.png");
```
MeniKadosh1: If it supports only RQM so why to do IF !=QC instead of IF == RQM?

uzanh: what about rally and other application, i'm thinking they should use by ID as well?

PR #89, opened 2018-08-23
adding binding between DataFile field and Param Name field with Param…

sandesh05:
… List

### DIFF
--- a/Ginger/Ginger/Actions/ActCreatePDFChartEditPage.xaml.cs
+++ b/Ginger/Ginger/Actions/ActCreatePDFChartEditPage.xaml.cs
@@ -23,6 +23,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Windows;
+using System.Windows.Controls;
 
 namespace Ginger.Actions
 {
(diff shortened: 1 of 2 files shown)


Ginger/Ginger/Actions/ActCreatePDFChartEditPage.xaml.cs
```diff
@@ -32,20 +33,27 @@ namespace Ginger.Actions
...
             DataFileNameTextBox.Init(mAct.GetOrCreateInputParam(ActCreatePDFChart.Fields.DataFileName), ActInputValue.Fields.Value);
-            ParamsComboBox.Init(mAct.GetOrCreateInputParam(ActCreatePDFChart.Fields.ParamName), SetParamsCombo(),true);
+            GingerCore.General.ObjFieldBinding(ParamsComboBox.ComboBox, ComboBox.SelectedValueProperty, mAct, ActCreatePDFChart.Fields.ParamName);
+            GingerCore.General.ObjFieldBinding(ParamsComboBox.ComboBox, ComboBox.ItemsSourceProperty, mAct, ActCreatePDFChart.Fields.ParamList);
+            DataFileNameTextBox.ValueTextBox.TextChanged += ValueTextBox_TextChanged;           
+        }
+
+        private void ValueTextBox_TextChanged(object sender, TextChangedEventArgs e)
+        {
+            SetParamsCombo();
```
JaspreetGIT: Add mAct.ParamList = SetParamsCombo here.

Also i Please remove the paramcombo init in Browse_click event

Ginger/Ginger/Actions/ActCreatePDFChartEditPage.xaml.cs
```diff
@@ -65,31 +73,38 @@ private void BrowseDataButton_Click(object sender, RoutedEventArgs e)
...
+                DataFileNameTextBox.ValueTextBox.Text = FileName;                
+            }           
         }
 
-        private List<string> SetParamsCombo()
+        private void SetParamsCombo()
         {
+            List<string> tmp=new List<string>();
             object tempFileName = mAct.GetDataFileName();
-            if (ReferenceEquals(tempFileName,null)) return new List<string>();
```
JaspreetGIT: mAct.ParamList will retain the old value in case TempFileName is null. So better keep the return type as List<string> as before. and change in the call as per comment in text change event.

PR #92, opened 2018-08-24
Bug fix/6303 java diver agent is not available

ravirk91:
Bug 6303 - java diver agent is not available


Ginger/GingerCore/Platforms/Agent.cs
```diff
@@ -235,6 +235,7 @@ public override bool FilterBy(eFilterBy filterType, object obj)
         #endregion
 
         public bool IsFailedToStart = false;
+        public bool isRunning = false;
```
JinendraGhodke: This flag is not needed

Ginger/GingerCore/Platforms/Agent.cs
```diff
@@ -692,6 +692,16 @@ public void Close()
                 OnPropertyChanged(Fields.IsWindowExplorerSupportReady);
             }
         }
+
+        public void ResetAgentStatus(eStatus status)
```
JinendraGhodke: setting isrunning false is not sufficient. Because agent is still running and we just setting flag to false. Need to close the agent as well

Ginger/GingerCore/Actions/ActAgentManipulation.cs
```diff
@@ -66,15 +66,25 @@ public override void Execute()
                     {
                         this.Error = "Agent not running";
                     }
-                   
-                    RunOnBusinessFlow.CurrentActivity.CurrentAgent.Close();
+                    if (RunOnBusinessFlow.CurrentActivity.CurrentAgent.Status == Agent.eStatus.FailedToStart)
+                    {
+                        RunOnBusinessFlow.CurrentActivity.CurrentAgent.ResetAgentStatus(RunOnBusinessFlow.CurrentActivity.CurrentAgent.Status);
+                        this.ExInfo = "Agent is not running, status is updated.";
```
JinendraGhodke: Looks good. Just change the message to "Agent is not running, failed to start status is reset."

PR #100, opened 2018-08-29
fixing save issue from solution tab

ravirk91:
Bug-6322- Sync issues between Solution tab and Automation tab


Ginger/Ginger/Repository/LocalRepository.cs
```diff
@@ -1265,6 +1265,7 @@ public void SaveAllSolutionDirtyItems(bool refreshSolutionTree = false)
                 Reporter.CloseGingerHelper();
                 return;
             }
+
```
JinendraGhodke: don't have any change in this file. Please exclude this

PR #108, opened 2018-09-01
Bugfix bdd issues

JaspreetGIT:
Handled Import Feature File issues.
Added a new page in Gherkin Wizard to select target folder
Added Folder select in Gherkin Page for Create business Flow


Ginger/Ginger/SolutionWindows/TreeViewItems/BusinessFlowsFolderTreeItem.cs
```diff
@@ -138,14 +139,19 @@ List<ITreeViewItem> ITreeViewItem.Childrens()
...
                 }
                 else
-                {                    
-                    BFs = App.LocalRepository.GetSolutionBusinessFlows(specificFolderPath: Path);
+                {   
                     AddsubFolders(Path, Childrens);
+                    if (mViewMode == eBusinessFlowsTreeViewMode.FoldersOnly)
+                        return Childrens;
+                    BFs = App.LocalRepository.GetSolutionBusinessFlows(specificFolderPath: Path);
                 }
```
MeniKadosh1: i don't think this the solution for Folder selection because this is specific handling only for BF's tree, what if we will want to select Run Set folder? we don't want to modify all Folders tree items.
solution should be on TreeSelectionPage which will get param to show only folders, this config will pass to UCTree which will avoid adding treeitems which are not Folders (you can use the isExpandable flag to know which is Folder) in generic way.

Ginger/Ginger/UserControlsLib/TextEditor/Gherkin/GherkinPage.xaml.cs
```diff
@@ -514,6 +510,25 @@ private void GeneratBFButton_Click(object sender, RoutedEventArgs e)
...
+                    bfsFolder = new BusinessFlowsFolderTreeItem(WorkSpace.Instance.SolutionRepository.GetRepositoryItemRootFolder<BusinessFlow>(), eBusinessFlowsTreeViewMode.FoldersOnly);
+                }
+                else
+                {
+                    bfsFolder = new BusinessFlowsFolderTreeItem(eBusinessFlowsTreeViewMode.FoldersOnly);//create new tree each time for now to allow refresh
+                }                 
+                bfsFolder.Folder = GingerDicser.GetTermResValue(eTermResKey.BusinessFlows);
+                bfsFolder.Path = App.UserProfile.Solution.BusinessFlowsMainFolder;
+                bfsFolder.IsGingerDefualtFolder = true;
+                SingleItemTreeViewSelectionPage mTargetFolderSelectionPage = new SingleItemTreeViewSelectionPage(GingerDicser.GetTermResValue(eTermResKey.BusinessFlows), eImageType.BusinessFlow, bfsFolder, SingleItemTreeViewSelectionPage.eItemSelectionType.Folder, true, SingleItemTreeViewSelectionPage.eItemEnableEventType.Select);
```
MeniKadosh1: Consider that me and Yaron working now on modifing all tree items including BF's tree items to work with new SolutionRepository, that will break above code, if you will push code fist so later please check it is still working

Ginger/Ginger/UserControlsLib/UCTreeView/SingleItemTreeViewSelectionPage.xaml.cs
```diff
@@ -178,5 +185,18 @@ private void Tree_ItemDoubleClick(object sender, EventArgs e)
...
+            if(itvItem.NodeObject() == null)
+                mSelectedItems.Add(itvItem);
+            else
+                mSelectedItems.Add(itvItem.NodeObject());
+            if (SelectCurrentItem())
+            {
+                OnSelectionDone(new SelectionTreeEventArgs(mSelectedItems));
+            }            
+        }
     }
```
MeniKadosh1: can you please elaborate what this code is for?

Ginger/Ginger/UserControlsLib/UCTreeView/UCTreeView.xaml.cs
```diff
@@ -37,6 +37,7 @@ public partial class UCTreeView : UserControl, IDragDrop
         public event EventHandler ItemDropped;
         public delegate void ItemDroppedEventHandler(DragInfo DI);
         public bool TreeItemDoubleClicked = false;
+        public bool TreeChildFolderOnly = false;
```
MeniKadosh1: is TreeChildFolderOnly is been used?

Ginger/Ginger/UserControlsLib/UCTreeView/SingleItemTreeViewSelectionPage.xaml.cs
JaspreetGIT: This is adding the selecting item to the selectionlist in case node object of selected item is null.
I need the path of the selected node from the calling class. Earlier we were  updating the list only on double click , i added it for select event

Ginger/Ginger/SolutionWindows/TreeViewItems/BusinessFlowsFolderTreeItem.cs
JaspreetGIT: i could have done that, but the issue is in case of heavy solutions,get all children will get all the child elements and then only we will need to filter the ones which are folder items, which i snot good solution performance wise.

Ginger/Ginger/UserControlsLib/UCTreeView/SingleItemTreeViewSelectionPage.xaml.cs
MeniKadosh1: But you want to select only single folder, no? so why list? the actual selection is done with the "Select" button which takes the current selectedItem, i guess i'm missing something...

Ginger/Ginger/SolutionWindows/TreeViewItems/BusinessFlowsFolderTreeItem.cs
JaspreetGIT: This option also thought, but the problem is ITreeViewItem.Childrens() is called when the tree is expanded and since ITreeViewItem is interface I cannot define enum/bool at this level. So had to add it in specific treeitem.

Ginger/Ginger/UserControlsLib/UCTreeView/SingleItemTreeViewSelectionPage.xaml.cs
JaspreetGIT: We do not have select button here. as we are not opening it as window. So just implemented the same we had in select click. Also, yes we only need single item selected, but then we will need to define another selection argument which will return object instead of List<object>. So i kept it as list and outside reading the first element from the list

Ginger/Ginger/UserControlsLib/UCTreeView/UCTreeView.xaml.cs
```diff
@@ -40,7 +40,8 @@ public partial class UCTreeView : UserControl, IDragDrop
         public event EventHandler ItemDoubleClick;
         public event EventHandler ItemDropped;
         public delegate void ItemDroppedEventHandler(DragInfo DI);
-        public bool TreeItemDoubleClicked = false;        
+        public bool TreeItemDoubleClicked = false;
+        public bool TreeChildFolderOnly = false;
```
MeniKadosh1: i see you added the TreeChildFolderOnly but didn't used it, you should use it inside SetTreeNodeItemChilds() so it will be generic without touching the original folder treeItems - but it's your call to decide what is better

PR #121, opened 2018-09-06
Fixing of 3 (6318, 6345, 6319) defects and few additional issues

OlegOrlovsky:
This not all defects fixed from list that you assigned to me, but fixed staff is enought critical.
Please consider this pull request for now. I will proceed to work on other defects after nearest vocation.


Ginger/Ginger/Actions/ActionConversion/ActionConverterPage.xaml.cs
```diff
@@ -315,6 +318,7 @@ private void btnConvertAction_Click(object sender, RoutedEventArgs e)
                         }
                     }
                 }
+                Mouse.OverrideCursor = null;
             }
```
MeniKadosh1: please do Mouse.OverrideCursor = null; inside Finally block so it will be done even if exception will happen in middle of code (so mouse won't get stuck with Processing icon).
for showing Process is running better to show our 'Processing" imageMaker image, see sample in POM wizard when we learn the object so we show the rotatating processing image to make clear we running in backend.

Ginger/GingerCore/Actions/ActGenElement.cs
```diff
@@ -166,6 +166,7 @@ Act IObsoleteAction.GetNewAction()
                     case eGenElementAction.Focus:
                         NewActUIElement.ElementAction = ActUIElement.eElementAction.SetFocus;
                         break;
+                    case eGenElementAction.SetAttributeUsingJs:
                     case eGenElementAction.RunJavaScript:
```
MeniKadosh1: we can't convert SetAttributeUsingJs to eElementAction.RunJavaScript because it has diffrent implementation, it won't pull the attribute value with UIElementAction.
leave it, no need to convert it.

MeniKadosh1:
Oleg please updated and then merge this branch

MeniKadosh1:
i merged it with master

PR #130, opened 2018-09-13
Fix for bug 6378 - API Model- Learn from XML- Params Counter not been reset

yulev55:


yulev55:
Yes,
For JSON the issue doesn't occur

PR #143, opened 2018-09-21
fix for failing recover for find and replace

KashmiraDeshpande:


Ginger/Ginger/SolutionAutoSaveAndRecover/RecoverPage.xaml.cs
```diff
@@ -111,8 +111,9 @@ private void RecoverButton_Click(object sender, RoutedEventArgs e)
                         ri.Status =  eRecoveredItemStatus.RecoveredFailed;
                         return;
                     }
-
-                    File.Replace(ri.RecoveredItemObject.FileName, originalItem.FileName,originalItem.FilePath + ".bac");
+                    File.Delete(originalItem.FileName);
+                    File.Move(ri.RecoveredItemObject.FileName, originalItem.FileName);
+
```
JinendraGhodke: We are not sure if replace is doing the issue and may be it is because of file watcher. 
PauseFileWatcher for the Repository folder before replace and then resume it after replace. See the relevant code for pause and resume its already used in few places.

PR #155, opened 2018-09-26
Bugfix/fixing of web convertor+window explorer issues

OlegOrlovsky:
Fixing of 6379 
2 possible fixes for 6342 - one to load all xpathes directly to grid (performance), second only by clicking on specific row (like was mentoined in defect description)


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -3453,6 +3453,7 @@ List<ElementInfo> IWindowExplorer.GetVisibleControls(List<eElementType> filtered
             Driver.SwitchTo().DefaultContent();
 
             list = GingerCore.General.ConvertObservableListToList<ElementInfo>((GetAllElementsFromPage("", filteredElementType, foundElementsList)));
+            // list.ForEach(z => z.XPath = GenerateXpathForIWebElement((IWebElement)z.ElementObject, "")); //  also can be fix for 6342 - to discuss
```
MeniKadosh1: This will have impact i belive, let's discuss it tomorrow

MeniKadosh1: just noticed this is a comment which was removed so avoid it

MeniKadosh1:
I merged it so you won't need to update it with latest again, please delete the branch if not needed anymore

PR #156, opened 2018-09-27
JavaDriver: Adding support for select element bymultipleproperty

ravirk91:
Java driver: Adding support for select element ByMultipleProperties


Ginger/GingerCore/Drivers/JavaDriverLib/GingerJavaAgent/agent/com/amdocs/ginger/XPathHelper.java
```diff
@@ -374,6 +375,28 @@ private int GetNameComponentIndex(Component comp)
...
+
+	public Component GetComponentByMultipleProperties(Window currentWindow, String multipleProperties)
+	{
+		if(currentWindow == null)
+		{
+			GingerAgent.WriteLog("\n Error: Window is not set.");
+			return null;
+		}
+		Container container =currentWindow;
+		String [] pathNodes=multipleProperties.split("/");
```
JinendraGhodke: In case of Multi property locate value will have only one node. So no need to split. Also a loop is not needed. Please update this

PR #163, opened 2018-10-04
Adding support for response type application/pdf

ravirk91:
Bug-6351-need a solution which can support application/pdf MIME type


ravirk91:
closing this pull request, as open another pull request for same fixes

PR #171, opened 2018-10-09
6445_WindowExplorer_Exception when clicking on grid_Fix adding support for Java filtering

EliranBarDavid:


Ginger/Ginger/WindowExplorer/Common/WindowExplorerPage.xaml.cs
```diff
@@ -1147,6 +1147,10 @@ private void SetAutoMapElementTypes()
                     WebPlatform webPlatformInfo = new WebPlatform();
                     UIElementsTypeList = webPlatformInfo.GetPlatformUIElementsType();
                     break;
+                case ePlatformType.Java:
+                    JavaPlatform javaPlatformInfo = new JavaPlatform();
+                    UIElementsTypeList = javaPlatformInfo.GetPlatformUIElementsType();
+                    break;
             }
```
MeniKadosh1: let's talk on this change, only changes related to Web supposed to be done to window explorer because only it supports POM now, not sure why we needed to do changes also to Java to work, what about Windows/Power builder application? they will need change as well?

EliranBarDavid: I corrected the code to support Web and avoid exception from the other platforms,
please check,
Thanks,

Ginger/Ginger/WindowExplorer/Common/WindowExplorerPage.xaml.cs
```diff
@@ -1148,10 +1148,12 @@ private void SetAutoMapElementTypes()
                     UIElementsTypeList = webPlatformInfo.GetPlatformUIElementsType();
                     break;
             }
-
-            foreach (eElementType eET in UIElementsTypeList)
+            if (UIElementsTypeList != null)
             {
```
MeniKadosh1: do not show filter page if FilteringCreteriaList is empty

PR #175, opened 2018-10-09
added directory separator to fix format of solution folder path

sudarshan0212:


Ginger/Ginger/Solution.cs
```diff
@@ -55,7 +55,7 @@ public static Solution LoadSolution(string solutionFileName, bool startDirtyTrac
             txt = txt.Replace("Ginger.Environments.Solution", "Ginger.SolutionGeneral.Solution");//changed the namespace so need to handle old xml's
             Solution solution = (Solution)NewRepositorySerializer.DeserializeFromText(txt);
             solution.FilePath = solutionFileName;
-            solution.Folder = Path.GetDirectoryName(solutionFileName);
+            solution.Folder = Path.GetDirectoryName(solutionFileName) + Path.DirectorySeparatorChar;
             if (startDirtyTracking)
```
MeniKadosh1: please also look for all places in which we use this solution.Folder for building path and use "Path.Combin()" so all code places won't be sensitive to missing/extra "/" 

notice that in my branch of fixes i already added few Path.Combin to Soluton.Folder

PR #182, opened 2018-10-10
Bugfix/6462 window platform unable to spy

JaspreetGIT:


Ginger/Ginger/WindowExplorer/Common/ControlActionsPage.xaml.cs
```diff
@@ -157,6 +158,11 @@ private void InitActionsGrid()
 
         private void AddToPageList()
         {
+            if (App.PageList == null)
```
MeniKadosh1: but why now it comes as null compared to before?

Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -309,111 +309,129 @@ public void ReportUnusedVariables(object obj, List<string> usedVariables)
...
+                    if (var.Name.Equals(((AnalyzeSolution)sender).ItemName))
                     {
-                        if (var.Name.Equals(AA.ItemName))
-                        {
-                            BusinessFlow.SolutionVariables.Remove(var);
-                            AA.Status = eStatus.Fixed;                                                                                    
-                        }
+                        BusinessFlow.SolutionVariables.Remove(var);
+                        ((AnalyzeSolution)sender).Status = eStatus.Fixed;
+                        break;
```
MeniKadosh1: why we have all of this analyze code inside AnalyzerPage? the analyze need to be inside the items itself like ActivityAnalyzer/ BusinessFlowAnalyzer/... what i'm missing?

Ginger/Ginger/WindowExplorer/Common/ControlActionsPage.xaml.cs
JaspreetGIT: Because It was being intialized in MainWindow.xaml earlier where we were holding Other Pages LIke SolutionPage/RunSetPage in teh PageList

PR #185, opened 2018-10-10
Bug fix/6441 after running a business flow a new is added at the end

ravirk91:
Bug fix/6441 after running a business flow a new is added at the end

### DIFF
--- a/Ginger/Ginger/App.xaml.cs
+++ b/Ginger/Ginger/App.xaml.cs
@@ -927,12 +927,14 @@ namespace Ginger
                 }
 
                 //load new Business Flow as default
+               // TODO: Need to check is actually below LoadDefaultBussinessFlow()method is required here
                 App.BusinessFlow = LoadDefaultBusinessFlow();
 
             }
             catch (Exception ex)
             {
                 Reporter.ToLog(eLogLevel.ERROR, "Failed to load the recent " + GingerDicser.GetTermResValue(eTermResKey.BusinessFlow) + " used", ex);
+                // TODO: Need to check is actually below LoadDefaultBussinessFlow()method is required here
                 LoadDefaultBusinessFlow();
             }
         }
--- a/Ginger/Ginger/App.xaml.cs
+++ b/Ginger/Ginger/App.xaml.cs
@@ -927,7 +927,6 @@ namespace Ginger
                 }
 
                 //load new Business Flow as default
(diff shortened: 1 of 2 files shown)


Ginger/Ginger/App.xaml.cs
```diff
@@ -927,12 +927,14 @@ private static void LoadRecentBusinessFlow()
                 }
 
                 //load new Business Flow as default
+               // TODO: Need to check is actually below LoadDefaultBussinessFlow()method is required here
                 App.BusinessFlow = LoadDefaultBusinessFlow();
```
MeniKadosh1: LoadRecentBusinessFlow() indeed needs to be changed now that we don't have Automate tab.
please change it as following: 
check if there are more than 0 BF's if yes so set App.BusinessFlow with first BF in list, if not having so use LoadDefaultBusinessFlow() to add defualt BF and set it to  App.BusinessFlow

Ginger/GingerCore/Repository/RepositorySerializer.cs
```diff
@@ -489,7 +489,8 @@ private static object xmlReadObject(Object Parent, XmlReader xdr, RepositoryItem
 
                 if (obj == null)
                 {
-                    throw new Exception("Error:Cannot create Class: " + ClassName);
+                      Reporter.ToLog(eLogLevel.INFO, "Error:Cannot create Class: " + ClassName, writeOnlyInDebugMode: true);
+                      return null;
```
MeniKadosh1: issue with this is that we might miss that there is some issue with some class, at least change the Log to be: eLogLevel.Error and remove this writeOnlyInDebugMode: true- it always needs to be added to Log

Ginger/Ginger/App.xaml.cs
ravirk91: it is already handled in code. 
image

Removing the TODO comment.

Ginger/GingerCore/Repository/RepositorySerializer.cs
ravirk91: changed the Log from info to error and removed writeOnlyInDebugMode: true.

MeniKadosh1:
Yaron is fixing the serilizer and i already fixing the LoadRecentBusinessFlow on diffrent branch so closing this branch and related ticket.

MeniKadosh1:
Ravi please delete the branch

PR #217, opened 2018-10-17
Fix to use ReposiotryItembase instead of RI in var dependancy and App…

JinendraGhodke:
Fix to use ReposiotryItembase instead of RI in var dependancy and Applist page


.gitignore
```diff
@@ -328,3 +328,4 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+/Ginger/GingerCore/Drivers/JavaDriverLib/GingerJavaAgent/.externalToolBuilders
```
MeniKadosh1: what is this gitIgnore file?

PR #223, opened 2018-10-18
Fix for Run flow without Analyzer from runset

KashmiraDeshpande:


JinendraGhodke:
Closing this pull request. Meni will create new Pull request with fixes for all Boolean field which needs to be true

PR #231, opened 2018-10-23
Bug fix/6493 analyzer not recognizing parameters

sudarshan0212:


Ginger/GingerCore/Variables/VariableBase.cs
```diff
@@ -345,6 +345,22 @@ public static void GetListOfUsedVariables(object item, ref List<string> usedVari
                                                 usedVariables.Add(value.ToString());
                                     }
                                 }
+                                else if(mi.Name == "ValueCalculated" && mi.DeclaringType.Name == "FlowControl")
```
JinendraGhodke: Add a comment explaining that this condition is added for Set variable flow control type

PR #236, opened 2018-10-24
Feature/6351 legacy rest action addinng support for pdf response

ravirk91:
Feature/6351 legacy rest action addinng support for pdf response

### DIFF
--- a/Ginger/GingerCore/Actions/Rest/ActREST.cs
+++ b/Ginger/GingerCore/Actions/Rest/ActREST.cs
@@ -45,7 +45,7 @@ namespace GingerCore.Actions.REST
 
         // static CookieCollection SessionCokkies = new CookieCollection();
         static Dictionary<string, Cookie> _sessionCokkiesDic = new Dictionary<string, Cookie>();
-      //  static Dictionary<string, CookieCollection> HostCookies = new Dictionary<string, CookieCollection>();
+        //  static Dictionary<string, CookieCollection> HostCookies = new Dictionary<string, CookieCollection>();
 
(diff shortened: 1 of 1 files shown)


Ginger/GingerCore/Actions/Rest/ActREST.cs
```diff
@@ -513,46 +515,87 @@ public override void Execute()
...
+
+                    String fileName = "";
+                    String timeStamp = DateTime.Now.ToString("dd_MM_yyyy_HH_mm_ss_fff");
+                    fileName += this.Description;
+                    fileName += "_" + "Response";
+                    fileName += "_" + timeStamp;
+
+                    string DirectoryPath = SaveRequestResponseFolderPath.ValueForDriver;
+                    if (DirectoryPath.StartsWith(@"~\"))
+                    {
```
JinendraGhodke: Create a function for checking directory path and reuse for XML as well as PDF

Ginger/GingerCore/Actions/Rest/ActREST.cs
```diff
@@ -513,46 +515,87 @@ public override void Execute()
...
+                        Directory.CreateDirectory(DirectoryPath);
+                    }
+                    string filePath = DirectoryPath + "//" + fileName + ".pdf";
+
+                    File.WriteAllBytes(filePath, data);
+
+                }
+                else
+                {
+                    //TODO: check if UTF8 is good for all
```
JinendraGhodke: We always keep the happy path scenarios first. So move the else part to If and vice versa

PR #243, opened 2018-10-27
Bugfix filewatcher fixes

yaronweiss:


Ginger/GingerTest/POMs/SingleItemTreeViewExplorerPagePOM.cs
```diff
@@ -206,7 +207,12 @@ public IEnumerable<TreeViewItem> GetRootItems()
...
-
+        /// <summary>
+        ///  Check if item exist in the tree, include retry mechanism of 10 sec
+        ///  Use it when you expect the item to be fiund
+        /// </summary>
+        /// <param name="title"></param>
+        /// <returns></returns>
         internal bool IsItemExist(string title)
         {            
             TreeViewItem tvi = FindItem(title);
```
MeniKadosh1: those FindItems functions are based on the item title? if yes it is not good because user may have the same item name numbe rof times, like same BF name on same or diffrent folders. need to do search by the tree item Object

PR #244, opened 2018-10-28
BugFix/6528-Adding support for async population on the grid

EliranBarDavid:


MeniKadosh1:
PR was pushed and branch was deleted.
please update related tickets and ask for retest.
also please rested well all API import types to make sure it works well with the Async approach.

PR #245, opened 2018-10-28
Bug fix/6215 export to qc failure failed to be exported to alm

uzanh:


Ginger/GingerCore/ALM/RQM/ImportFromRQM.cs
```diff
@@ -840,7 +840,11 @@ private static void SaveItemFields(ObservableList<ExternalItemFieldBase> refresh
         public static ObservableList<ExternalItemFieldBase> GetOnlineItemFields(BackgroundWorker bw)
         {
             ObservableList<ExternalItemFieldBase> fields = new ObservableList<ExternalItemFieldBase>();
-
+            if(bw == null)
+            {
+                bw = new BackgroundWorker();
+                bw.WorkerReportsProgress = true;
+            }
             //TODO : receive as parameters:
```
MeniKadosh1: in which cases the BackgroundWorker comes null?

Ginger/Ginger/ALM/ALMConnectionPage.xaml.cs
```diff
@@ -410,6 +410,11 @@ private void StyleRadioButtons()
         private void SaveALMConfigs()
         {
             ALMIntegration.Instance.SyncConfigurations();
+            Ginger.SolutionGeneral.Solution lastSavedSolution = Ginger.SolutionGeneral.Solution.LoadSolution(App.UserProfile.Solution.FilePath, false);
+            if (!(App.UserProfile.Solution.ALMServerURL == lastSavedSolution.ALMServerURL && App.UserProfile.Solution.ALMDomain == lastSavedSolution.ALMDomain && App.UserProfile.Solution.ALMProject == lastSavedSolution.ALMProject))
+            {
+                App.UserProfile.Solution.ExternalItemsFields.Clear();
+            }
             App.UserProfile.SaveUserProfile();
```
MeniKadosh1: we agreed this approch won't work no? 
because what if:
1- user set QC A + fields and close Ginger
2- open Ginger shift to QC B and load it fields
when he will do save above code will clear all of his QC B fields which were loaded in valid way.
i think we should clear the fields on cache as soon as user changes the URL/domain/Project and not on save so if user didn't loaded new fields he will have empty list on save and if he handeld new list so we will save the valid list.

Ginger/GingerCore/ALM/RQM/ImportFromRQM.cs
uzanh: Export BF & AG and Items field configuration.

PR #255, opened 2018-10-31
Bug fix/6571 correct spelling mistakes

sudarshan0212:


Ginger/Ginger/UserControlsLib/PieChart/StatusColorSelector.cs
```diff
@@ -71,7 +71,7 @@ public Brush SelectBrush(object item, int index)
                 case "Started":
                 case "Running":
                 case "Wait":
-                case "Cancelling":
+                case "Canceling":
                     return (Brush)resourceDictionary["$RunningStatusColor"];
```
MeniKadosh1: is this "Canceling" will be same as what will come from ENUM value or other?

sudarshan0212: Yes right it is coming from Enum, need to change enum value as well.

PR #268, opened 2018-11-01
Added code to sort the parameters while exporting to datasource

shahanemahesh:


shahanemahesh:
I tried to check this option first with Jinendra and then tried with Yaron as well he was ok with the change i implemented, currently there is no such option in the ucGrid, please let me know the option if want to change

JaspreetGIT:
I have checked updated code.. its not the correct implementation. Discuss with meni about why Grid items are getting sorted when launch API model page

PR #281, opened 2018-11-04
adding licence header to remaining files

shahanemahesh:


shahanemahesh:
As suggested changing base branch to Master from Release3.0

PR #328, opened 2018-11-12
Nuget Updates for Selenium

ambarkhan:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -435,8 +434,8 @@ public override void StartDriver()
                         break;
 
                     case eBrowserType.PhantomJS:
-                        string PhantomJSServerPath = Path.Combine(General.GetGingerEXEPath(), @"Drivers\PhantomJS");
-                        Driver = new PhantomJSDriver(PhantomJSServerPath);
+
+                        throw new NotSupportedException("Support for PhantomJS is ended");
                         break;
```
MeniKadosh1: so we need also to remove the "SelenuimPhantomeJS" driver option from Web Agent Drivers.
after removing check what happens if you had Agent mapped to it in previous Ginger version and if you load it with latest. expected is that Log will show error but solution will be loaded without this specific Agent.

PR #350, opened 2018-11-18
Feature plugin driver

yaronweiss:


Ginger/GingerCore/Platforms/Agent.cs
```diff
@@ -417,45 +447,103 @@ public void StartDriver()
...
             }
+
+            Stopwatch st = Stopwatch.StartNew();
+            while (mGingerNodeInfo == null && st.ElapsedMilliseconds < 30000) // max 30 seconds to wait
+            {
+
+                //!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+
+                mGingerNodeInfo = (from x in WorkSpace.Instance.LocalGingerGrid.NodeList where x.ServiceId == "SeleniumChromeDriver" select x).FirstOrDefault();  // Keep First!!!
+                if (mGingerNodeInfo != null) break;
```
MeniKadosh1: why "SeleniumChromeDriver" is hard coded here?

Ginger/GingerCore/Platforms/TargetApplication.cs
```diff
@@ -63,5 +63,11 @@ public override string ItemName
                 return;
             }
         }
+
+        [IsSerializedForLocalRepository]
+        public Agent.eAgentType TargetAgentType { get; set; }
+
```
MeniKadosh1: how TargetAgentType is related to TargetApplication? we said we will remove the dependncy on the Target Application and the plugins dependency will be added to the Runner based on added Actions no?

PR #367, opened 2018-11-22
Bug fix/6671 invalid operation exception while executing unix regression suite

JaspreetGIT:


MeniKadosh1:
Need to be pushed to Master and not to Beta

PR #387, opened 2018-11-29
Test/variables

ravikirankethe:
Meni, 

Fixed below issues for variables and can you please review this and merge to the master. I tested all variables UnitTC's and it worked fine. 

- Added couple of TC's
- Fixed Type
- Replaced Fields usage with NameOf from VariableBase


Ginger/GingerCoreCommon/VariablesLib/VariableBase.cs
```diff
@@ -217,12 +194,12 @@ public string Formula
...
         private bool mVarValChanged;
         [IsSerializedForLocalRepository]
-        public bool VarValChanged { get { return mVarValChanged; } set { mVarValChanged = value; OnPropertyChanged(Fields.VarValChanged); } }
+        public bool VarValChanged { get { return mVarValChanged; } set { mVarValChanged = value; OnPropertyChanged(nameof(VarValChanged)); } }
 
         //used to identify the variables which the user customized for specific BF run
         private bool mDiffrentFromOrigin;
         [IsSerializedForLocalRepository]
-        public bool DiffrentFromOrigin
+        public bool DifferentFromOrigin
```
MeniKadosh1: You can't change name of fields which been serilized to XML because then previous saved XML's will have old name of it and field unserilize will fail because it won't found the matching field for it.

i see you doing a lot of typo fixing but please do not do it to fields which has the  [IsSerializedForLocalRepository] attribute

Ginger/GingerCoreCommon/VariablesLib/VariablePasswordString.cs
```diff
@@ -64,7 +64,10 @@ public override void GenerateAutoValue()
 
         public override List<VariableBase.eSetValueOptions> GetSupportedOperations()
         {
-            throw new System.NotImplementedException();
+            List<VariableBase.eSetValueOptions> supportedOperations = new List<VariableBase.eSetValueOptions>();
```
MeniKadosh1: why this was added?
Password variable value can't be set during run time or been reset. user need to set it offline as part of flow design.

ravikirankethe: Earlier it was not there and it is pulling all the string operations to this one which password is not supporting. When I was writing TC's it is showing ClearSpecialChar as one of the supported operation and PasswordString doesn't have that implementation. So I added this validation to ensure that it shows only the supported operations.

Ginger/GingerCoreCommon/VariablesLib/VariableBase.cs
MeniKadosh1: only this one, did you changed any other serialized field name?

Ginger/GingerCoreCommon/VariablesLib/VariablePasswordString.cs
MeniKadosh1: Those operation are shown as possible operation in SetVariableValue action, right? so after this addition those "SetValue" and "ResetValue" will appear as possible operations in the Action and they not supposed to be shown.
let me know what i'm missing

ravikirankethe: Reverted back to throw NotImplementedException exception

ravikirankethe:
@MeniKadosh1, Fixed the changes. It can be further reviewed and can be pushed if you are okay.

MeniKadosh1:
approved it, please push it once all Appvayer checks pass

MeniKadosh1:
merged and deleted branch

PR #388, opened 2018-12-02
Fix update steps while export BF or AG to ALM ota api

uzanh:


MeniKadosh1:
Fix was merged and branch was deleted

PR #415, opened 2018-12-11
fix filter issue, UI element refresh when changing element issue, POM…

EliranBarDavid:
… messages typo issues, removing locate by class auto leaned locator as it is not working most of the time


Ginger/Ginger/ApplicationModelsLib/POMModels/PomAllElementsPage.xaml.cs
```diff
@@ -144,7 +144,7 @@ private void CreateNewElemetClicked(object sender, RoutedEventArgs e)
         }
 
 
-        System.Windows.Threading.DispatcherTimer dispatcherTimer = null;
+        public static System.Windows.Threading.DispatcherTimer mDispatcherTimer = null;
```
MeniKadosh1: 1- why mDispatcherTimer is static? thoretically users can work on 2 POM pages in tree in parallel if each is connected to diffrent Agent. 
2- mDispatcherTimer  should not be Public and be touched from outside, instead expose "StopSpy()" public method which internally will make it False if needed

Ginger/GingerCore/Reporter/UserMessagesPool.cs
```diff
@@ -519,10 +519,10 @@ public static void LoadUserMessgaesPool()
...
+            Reporter.UserMessagesPool.Add(eUserMsgKeys.WizardCantMoveNextWhileInProcess, new UserMessage(eAppReporterMessageType.WARN, "Process is still running", "Ginger cannot move next until the process will be finished or stopped." + Environment.NewLine + "Please wait for the process to be finished or stop it and then click next.", MessageBoxButton.OK, MessageBoxResult.None));
+            Reporter.UserMessagesPool.Add(eUserMsgKeys.WizardCantMovePrevWhileInProcess, new UserMessage(eAppReporterMessageType.WARN, "Process is still running", "Ginger cannot move previous until the process will be finished or stopped." + Environment.NewLine + "Please wait for the process to be finished or stop it and then click previous.", MessageBoxButton.OK, MessageBoxResult.None));
+            Reporter.UserMessagesPool.Add(eUserMsgKeys.WizardCantFinishWhileInProcess, new UserMessage(eAppReporterMessageType.WARN, "Process is still running", "Ginger cannot move finish wizard until the process will be finished or stopped." + Environment.NewLine + "Please wait for the process to be finished or stop it and then click finish.", MessageBoxButton.OK, MessageBoxResult.None));
```
MeniKadosh1: 1. why 3 same UserMessages? make the next/Previous/finish word as parameter {0} and then set it in calling function based on scenario 
2. the header should be with capital letters like: "Process is Still Running"
3. no need to say "Ginger..." just: "Move Next operation is blocked till process will be finished..."

Ginger/Ginger/ApplicationModelsLib/POMModels/PomElementsPage.xaml.cs
```diff
@@ -311,6 +311,8 @@ private void SetLocatorsGridView()
             defView.GridColsView = new ObservableList<GridColView>();
             defView.GridColsView.Add(new GridColView() { Field = nameof(ElementLocator.Active), WidthWeight = 8, MaxWidth = 50, HorizontalAlignment = System.Windows.HorizontalAlignment.Center, StyleType = GridColView.eGridColStyleType.CheckBox });
             List<GingerCore.General.ComboEnumItem> locateByList = GingerCore.General.GetEnumValuesForCombo(typeof(eLocateBy));
+            locateByList.Remove(locateByList.Where(x=>((eLocateBy)x.Value) == eLocateBy.POMElement).FirstOrDefault());
```
MeniKadosh1: if .FirstOrDefault() will return null so then you code will try to Remove(null) and exception will throw, need to check first if not null and then do Remove

PR #440, opened 2018-12-25
adding path to POM adding click capability for span

EliranBarDavid:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -6327,58 +6336,67 @@ ElementInfo IXPath.UseRootElement()
...
-            object parentElement;
+        {
+            HtmlNode parentElementHtmlNode;
+            IWebElement parentElementIWebElement;
             object childElement = ElementInfo.ElementObject;
 
-            if(childElement == null)
-                Driver.FindElement(By.XPath(ElementInfo.XPath));
+            parentElementHtmlNode = ((HtmlNode)childElement).ParentNode;
+            parentElementIWebElement = ((IWebElement)childElement).FindElement(By.XPath(".."));
```
MeniKadosh1: Fix  parentElementHtmlNode = ((HtmlNode)childElement).ParentNode; to work correctly with "HTMLElementObject"
need to check if should do only one operation instead always both.

check if need to include the ElementInfo parentEI = xxx to HTMLElementObject is null checks

MeniKadosh1:
merged and deleted branch

PR #442, opened 2018-12-26
Feature/6720 headless chrome browser support

vyasprateek:
Original : 6711 (and 6720)
+Headless Browser : Supported by Chrome & Firefox
+Launch Browser Minimized : Supported by Chrome, Firefox & Internet Explorer
+Hide Browser Console (Command Prompt) Window


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -106,6 +106,21 @@ public enum eBrowserType
...
 
+        [UserConfigured]
+        [UserConfiguredDefault("false")]
+        [UserConfiguredDescription("Only for Chrome & Firefox | Set \"true\" to run the browser in background (headless mode) for faster Execution")]
+        public bool HeadlessBrowserMode { get; set; }
+
+        [UserConfigured]
+        [UserConfiguredDefault("false")]
+        [UserConfiguredDescription("Only for IE, Chrome & Firefox | Set \"true\" to Launch the Browser minimized")]
+        public bool BrowserMinimized { get; set; }
```
MeniKadosh1: seems like BrowserMinimized is relevent for all browser types (Edge as well) so no need to mention "Only for IE, Chrome,..."

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -106,6 +106,21 @@ public enum eBrowserType
...
+
+        [UserConfigured]
+        [UserConfiguredDefault("false")]
```
MeniKadosh1: defualt value for HideConsoleWindow should be "true" we don't want the console as defualt

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -314,23 +331,32 @@ public override void StartDriver()
...
                         else
                         {
-                            if (Convert.ToInt32(HttpServerTimeOut) > 60)
+                            if (Convert.ToInt32(HttpServerTimeOut) >= 60)
                             {
                                 InternetExplorerDriverService service = InternetExplorerDriverService.CreateDefaultService();
+
+                                if (HideConsoleWindow == true)
+                                    service.HideCommandPromptWindow = true;
+
```
MeniKadosh1: few issues here:
1- there is no handling in case the IE type is 64bit- see row 323 If block
2- because you set the HttpServerTimeOut condition to be >= it means that you modified the defualt handling because till now it was considered only if it was bigger than the defualt 60
3- in case user set the timeout to let's say 30 and set HideConsoleWindow to "true" then current code won't do HideConsoleWindow because main If will be false

you should modify the If to be: if (Convert.ToInt32(HttpServerTimeOut) != 60  || HideConsoleWindow == true)
and inside If block to handle either one of them or both of them as required

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -372,29 +398,36 @@ public override void StartDriver()
...
-                       
-                        if (Convert.ToInt32(HttpServerTimeOut) > 60)
+
+                        if (Convert.ToInt32(HttpServerTimeOut) >= 60)
                         {
                             FirefoxDriverService service = FirefoxDriverService.CreateDefaultService();
+
+                            if (HideConsoleWindow == true)
+                                service.HideCommandPromptWindow = true;
+
```
MeniKadosh1: same issues like for IE- see above comment

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -410,17 +443,25 @@ public override void StartDriver()
...
                         if (SeleniumUserArgs != null)
                             foreach (string arg in SeleniumUserArgs)
                                 options.AddArgument(arg);
-                        if (Convert.ToInt32(HttpServerTimeOut) > 60)
+                        if (Convert.ToInt32(HttpServerTimeOut) >= 60)
                         {
                             ChromeDriverService service = ChromeDriverService.CreateDefaultService();
+
+                            if (HideConsoleWindow == true)
+                                service.HideCommandPromptWindow = true;
```
MeniKadosh1: same issues like for IE- see above comment

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -429,25 +470,39 @@ public override void StartDriver()
...
+
+                    #region EDGE
                     case eBrowserType.Edge:
-                        if (Convert.ToInt32(HttpServerTimeOut) > 60)
+                        if (Convert.ToInt32(HttpServerTimeOut) >= 60)
                         {
                             EdgeDriverService service = EdgeDriverService.CreateDefaultService();
+
+                            if (HideConsoleWindow == true)
+                                service.HideCommandPromptWindow = true;
```
MeniKadosh1: same issues like for IE- see above comment

also i don't see here handling for HttpServerTimeOut so please add it

Ginger/GingerCore/Drivers/SeleniumDriver.cs
vyasprateek: Yes, it is relevant to all but it doesn't work for 'Edge'.

MeniKadosh1: even if it is doesn't work for Edge, remove the Only for...

Ginger/GingerCore/Drivers/SeleniumDriver.cs
MeniKadosh1: i see defualt value is still "false"

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -502,8 +557,12 @@ public override void StartDriver()
 
                             break;
                         }
+                        #endregion
                 }
 
+                if (BrowserMinimized == true && mBrowserTpe != eBrowserType.Edge)
```
vyasprateek: There's an Exception here as, "Exception in start driver OpenQA.Selenium.WebDriverException: Unexpected error. Unknown command received". Thus added a check to ignore calling Window.Minimze() for Edge Browser.

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -502,8 +557,12 @@ public override void StartDriver()
...
+                    Driver.Manage().Window.Minimize();
```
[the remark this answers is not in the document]
vyasprateek: Yes meni, an exception occurs when we try to minimize 'edge' in this way as, "Exception in start driver OpenQA.Selenium.WebDriverException: Unexpected error. Unknown command received"

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -307,30 +324,44 @@ public override void StartDriver()
                         {
                             string IEdriver64bitpath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Locati‌​on) + @"\Drivers\IE64BitDriver");
 
+                            InternetExplorerDriverService IEService = InternetExplorerDriverService.CreateDefaultService(IEdriver64bitpath);
+                            IEService.HideCommandPromptWindow = (HideConsoleWindow == true);
```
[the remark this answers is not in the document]
vyasprateek: No it isn't required. I forgot to change under IE part. I'll make the changes and will push the same

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -307,30 +324,44 @@ public override void StartDriver()
...
+
                             if (Convert.ToInt32(HttpServerTimeOut) > 60)
                             {
-                                InternetExplorerDriverService service = InternetExplorerDriverService.CreateDefaultService();
-                                Driver = new InternetExplorerDriver(service, ieoptions, TimeSpan.FromSeconds(Convert.ToInt32(HttpServerTimeOut)));
+                                Driver = new InternetExplorerDriver(IEService, ieoptions, TimeSpan.FromSeconds(Convert.ToInt32(HttpServerTimeOut)));
                             }
                             else
                             {
```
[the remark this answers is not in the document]
vyasprateek: I did not changed with a thought of it being an existing working code so was just unsure if there have had been any specific requirement to keep it like this.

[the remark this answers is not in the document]
vyasprateek: I discussed with Jinendra regarding HTTPServerTimeOut, he told that IE already being slow, has 60 seconds as default value to HTTPServerTimeOut and we've also mentioned that the Default/Recommended value for HttpServer Timeout for Web Action Completion is minimum 60 secs.

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -303,34 +320,39 @@ public override void StartDriver()
...
+                        {
+                            IEService = InternetExplorerDriverService.CreateDefaultService();
+                        }
 
-                            if (Convert.ToInt32(HttpServerTimeOut) > 60)
-                                Driver = new InternetExplorerDriver(IEdriver64bitpath, ieoptions, TimeSpan.FromSeconds(Convert.ToInt32(HttpServerTimeOut)));
-                            else
-                                Driver = new InternetExplorerDriver(IEdriver64bitpath, ieoptions);
+                        IEService.HideCommandPromptWindow = HideConsoleWindow;
+                        if (Convert.ToInt32(HttpServerTimeOut) > 60)
```
MeniKadosh1: why only if bigger than 60? why not to check if != 60?
*this question/comment is for all HttpServerTimeOut IF's

PR #448, opened 2018-12-31
Fixing the path issue on item usage page+ Adding generic function to …

JinendraGhodke:
Fixing the path issue on item usage page+ Adding generic function to check if Folder to avoid


JinendraGhodke:
Closing due to invalid Branch name

PR #449, opened 2018-12-31
Fixing the path issue on item usage page+ Adding generic function to …

JinendraGhodke:
Fixing the path issue on item usage page+ Adding generic function to check if Folder to avoid


JinendraGhodke:
Done the enhancement to skip only the relevant folder. now if someone will create Recover folder under business flow we will load it.

MeniKadosh1:
merged and deleted the branch

PR #456, opened 2019-01-02
Bug fix/6736 a task was canceled exception

ravirk91:
Fixing Task Canceled exception on ginger window closed for dos and unix driver


Ginger/GingerCore/Drivers/ConsoleDriverLib/DOSConsoleDriver.cs
```diff
@@ -97,7 +113,7 @@ public override string ConsoleWindowTitle()
 
         public override bool IsBusy()
```
JinendraGhodke: IsBusy is not used anywhere in the code. We can remove it.

MeniKadosh1:
branch was merged and deleted

PR #461, opened 2019-01-03
JavaDriver: Handling exception when tree item unalbe to get child node

ravirk91:
Bug 6409:Throwing exception when open window explorer and enter search item in search box


Ginger/Ginger/WindowExplorer/Java/JavaElementTreeItemBase.cs
```diff
@@ -34,17 +35,25 @@ public abstract class JavaElementTreeItemBase : TreeViewItemBase
...
-            List<PayLoad> controls = Response.GetListPayLoad();
-            List<ITreeViewItem> Childrens= new List<ITreeViewItem>();
-            if (Response.Name == "ContainerComponents")           
-               Childrens = GetControlsAsTreeItems(controls);
-          
-           if (Response.Name == "HTML Element Children")
-               Childrens = GetHTMLControlsAsTreeItems(controls);
+        {
+            List<ITreeViewItem> Childrens = new List<ITreeViewItem>();
+            try
```
JinendraGhodke: try catch is fine. But do we know why we are getting the exception in the first place ?

ravirk91: pushed changes for java side to avoid the exception

PR #465, opened 2019-01-04
Bug fix/6730 import from db unable to connect to oracle database

vyasprateek:
#6730 : Import from DB unable to connect to oracle database fixed. Handled : Database Connection String when entered in TNS/Host address field won't throw exception anymore and will handle & manipulate the string accordingly


MeniKadosh1:
Branch was merged and deleted

PR #469, opened 2019-01-04
Test/vb script tests

ravikirankethe:
@yaronweiss, added additional TC's. Please review and suggest me if we need to add more or can be merged with Master.


MeniKadosh1:
branch was merged and deleted

PR #471, opened 2019-01-06
Test/Unit test for pom learning

EliranBarDavid:


Ginger/GingerTest/APIModelLib/POMsTest.cs
```diff
@@ -0,0 +1,136 @@
...
+            //Assert  
+            Assert.AreEqual(mLearnedPOM.MappedUIElements.Count, 42, "POM.MappedUIElements.Coun check");
+            Assert.AreEqual(mLearnedPOM.UnMappedUIElements.Count, 77, "POM.UnMappedUIElements.Count check");
+            Assert.IsNotNull(EI1, "POM.Element learned check");
+            Assert.IsNotNull(EI2, "POM.Element learned check");
+        }
+
+        [TestMethod]
+        public void ValidateElementsProperties()
+        {
```
MeniKadosh1: not only count, validate full list of Properties is correct (name, class, x, y,... what ever will pull today for 2 elements, 1 button and 1 drop list)

Ginger/GingerTest/APIModelLib/POMsTest.cs
```diff
@@ -0,0 +1,136 @@
...
+            //Assert  
+            Assert.AreEqual(mLearnedPOM.MappedUIElements[0].Properties.Count, 8, "POM.properties check");
+            Assert.AreEqual(mLearnedPOM.UnMappedUIElements[1].Properties.Count, 7, "POM.properties check");
+        }
+
+
+        [TestMethod]
+        public void ValidateElementsLocators()
+        {
+            //Assert
```
MeniKadosh1: not only count, pelease check full list on locators that we fin for 1 button and 1 drop list and validate the xPath, RleativeXpath, Id, name, all we have on those elements

Ginger/GingerTest/APIModelLib/POMsTest.cs
```diff
@@ -0,0 +1,215 @@
...
+        [TestMethod]
+        public void ValidateElementsProperties()
+        {
+            //Act  
+            ElementInfo ButtonEI = mLearnedPOM.MappedUIElements.Where(x => x.ElementTypeEnum == eElementType.Button).FirstOrDefault();
+            ElementInfo ComboBoxEI = mLearnedPOM.MappedUIElements.Where(x => x.ElementTypeEnum == eElementType.ComboBox).FirstOrDefault();
+
+            //Assert  
+            Assert.AreEqual(ButtonEI.Properties.Count, 13, "POM.properties check");
+            Assert.AreEqual(ButtonEI.Properties[0].Name, "Platform Element Type", "POM.property 0 check");
```
MeniKadosh1: coverage is good now, but we should not do Index based testing- this is very breakable and will require annoying maintainance (any change will require updating all index numbers), instead, add small function to test class like "IsPropertyExist(string PropName, string PropValue) which will do Linq check and will return true/false if the property exist on no

Ginger/GingerTest/APIModelLib/POMsTest.cs
```diff
@@ -0,0 +1,215 @@
...
+        [TestMethod]
+        public void ValidateElementsLocators()
+        {
+            //Act
+            ElementInfo ButtonEI = mLearnedPOM.MappedUIElements.Where(x => x.ElementTypeEnum == eElementType.Button).FirstOrDefault();
+            ElementInfo ComboBoxEI = mLearnedPOM.MappedUIElements.Where(x => x.ElementTypeEnum == eElementType.ComboBox).FirstOrDefault();
+
+            //Assert
+            Assert.AreEqual(ButtonEI.Locators.Count, 4, "POM.Locators check");
+            Assert.AreEqual(ButtonEI.Locators[0].LocateBy, eLocateBy.ByID, "POM.Locator 0 .LocateBy check");
```
MeniKadosh1: same like properties, let's avoid using Index based comparison

MeniKadosh1:
merged and deleted branch

PR #531, opened 2019-01-18
Adding Unit Test for Soap UI wrapper Action

ravirk91:
Adding Unit Test for Soap UI wrapper Action

### DIFF
--- a/Ginger/Unit Tests/NonUITests/WebServicesTest.cs
+++ b/Ginger/Unit Tests/NonUITests/WebServicesTest.cs
@@ -23,6 +23,7 @@ using Amdocs.Ginger.Common.InterfacesLib;
 using Amdocs.Ginger.Repository;
 using Ginger.Run;
 using GingerCore;
+using GingerCore.Actions;
 using GingerCore.Actions.WebServices;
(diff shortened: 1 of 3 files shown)


Ginger/Unit Tests/NonUITests/WebServicesTest.cs
```diff
@@ -398,7 +399,50 @@ private string getXML()
...
+
+            ActWebAPISoap actSoapUi = new ActWebAPISoap();
+
+            var xmlFilePath = TestResources.GetTestResourcesFile(@"XML\calculator_soapui_project.xml");
+
+            actSoapUi.AddOrUpdateInputParamValue(ActSoapUI.Fields.ImportFile, xmlFilePath);
+
+            mBF.Activities[0].Acts.Add(actSoapUi);
+
+            if (actSoapUi.ActInputValues.Count > 0)
```
JinendraGhodke: we usually do not use foreach inside unit test.  Also the assert is not clear, there is if condition to check contains and then you doing assert. So it will always be true.

We should do assert on something like , Inputvalues count or Properties count etc.

PR #532, opened 2019-01-18
Feature/Perfecto add security token support

EliranBarDavid:


Ginger/GingerCore/Drivers/Mobile/Perfecto/PerfectoDriver.cs
```diff
@@ -140,7 +145,7 @@ public Boolean ConnectToPerfecto()
             }
             catch (Exception e)
             {
-                Reporter.ToLog(eAppReporterLogLevel.ERROR, "Error, Could not create Perfecto Mobile Automation Driver, " + e.Message);
+                Reporter.ToUser(eUserMsgKeys.FailedToConnectAgent, "Perfecto", e.Message);
             }
```
MeniKadosh1: we do not supposed to show User pop up messages from Drivers, what if this will run on remote machine, who will see the pop up? 
you need to write to log and if needed return flag saying Driver load failed.

PR #533, opened 2019-01-18
Added the code to handle the Update of DataSource from OutPut Values screen in edit action

shahanemahesh:


Ginger/GingerCoreNET/ValueExpressionLib/ValueExpression.cs
```diff
@@ -135,6 +136,12 @@ public string ValueCalculated
 
         public string Calculate(string expression)
         {
+            if (!string.IsNullOrEmpty(expression) && expression.StartsWith(DATA_SOURCE_PRE_FIX))
```
JaspreetGIT: This might cause issue when using Data Source in Value expression directly to get value or something.
Please test it once . Also test for all Data Source actions

PR #534, opened 2019-01-18
Feature/6842 pom browser action add go to from pom

vyasprateek:


Ginger/Ginger/Actions/ActBrowserElementEditPage.xaml.cs
```diff
@@ -100,8 +104,19 @@ private void SetVisibleControlsForAction()
                     if (mAct.ControlAction == ActBrowserElement.eControlAction.GotoURL)
                     {
                         xOpenURLInPnl.Visibility = System.Windows.Visibility.Visible;
+                        xURLSrcPnl.Visibility = System.Windows.Visibility.Visible;
```
MeniKadosh1: for testing that below code didn't broke existing Browser actions:
1- use last Ginger Beta or previous
2- create flow with few Browser actions which has GoToURL operation with few data combinatoins, also add Browser actions with other operations except GoToURL
3- save the solution
4- load the solution with your branch code and confirm that all previously created action loaded correctly and executed as expeted

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -405,8 +405,7 @@ public override void StartDriver()
                     case eBrowserType.Chrome:
                         ChromeOptions options = new ChromeOptions();
 
-                        if (BrowserMinimized == false)
-                            options.AddArgument("--start-maximized");
+                        options.AddArgument("--start-maximized");
```
MeniKadosh1: why BrowserMinimized was removed here?

Ginger/GingerCore/Actions/ActBrowserElement.cs
```diff
@@ -176,5 +185,18 @@ public override String ActionType
             }
 
             public override System.Drawing.Image Image { get { return Resources.ASCF16x16; } }
+
+        public string ElementLocateValue
+        {
```
MeniKadosh1: how "ElementLocateValue" is related?

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -5441,7 +5440,25 @@ public void ActBrowserElementHandler(ActBrowserElement act)
 
                 case ActBrowserElement.eControlAction.OpenURLNewTab:
                     OpenUrlNewTab();
-                    GotoURL(act, act.GetInputParamCalculatedValue("Value"));
+
+                    if ((act.GetInputParamValue(ActBrowserElement.Fields.URLSrc) == ActBrowserElement.eURLSrc.UrlPOM.ToString()))
+                    {
+                        string POMGuid = act.GetInputParamCalculatedValue("UrlPOM");
```
MeniKadosh1: you should never use strings ad ID's, this will cause very breakable code whic compilr can't catch.
"UrlPOM" (which better be called "PomGUID") need to be added as to actBrowserAction.Fields and then refer here to the Field and not use string

Ginger/GingerCore/Drivers/SeleniumDriver.cs
vyasprateek: It's removed because many of the Web applications have different css properties for different views and maximized browser exposes each & every element that exist on the page. On the other hand, '_BrowserMinized_' property just minimizes the browser once it is launched, thus, both the actions occurs one after the other as,adding argument "--start-maximized" launches the browser in maximized window, once finished, browser will be minimized if '_BrowserMinized_' is set to '_true_'.

PR #557, opened 2019-01-23
Added the code to handle the exception

shahanemahesh:


Ginger/Ginger/Actions/UCValueExpression.xaml.cs
```diff
@@ -132,6 +134,7 @@ private void OpenExpressionEditorButton_Click(object sender, RoutedEventArgs e)
         {
             ValueExpressionEditorPage w = new ValueExpressionEditorPage(obj, AttrName, HideVEBusinessFlowAndActivityVariables);
             w.ShowAsWindow();
+            ValueTextBox.Text = Convert.ToString(obj.GetType().GetProperty(AttrName).GetValue(obj));
         }
```
MeniKadosh1: i don't think thst fix for this bug should be on the generic UCValueExpression, please ping me on Monday to talk on it

PR #566, opened 2019-01-24
Fixing SeleniumDriver GenerateXpathForIWebElement function to be align with the xpath from HTMLNode

EliranBarDavid:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -4679,7 +4679,7 @@ public ElementInfo GetHTMLElementInfoFromIWebElement(IWebElement EL, string path
         public string GenerateXpathForIWebElement(IWebElement IWE, string current)
         {
             if (IWE.TagName == "html")
-                return "/" + IWE.TagName + current;
+                return "/" + IWE.TagName +"[1]" + current;
```
MeniKadosh1: not sure what is issue and solution here but counting on you...

EliranBarDavid: We need to include all indexes in the Xpath in order to make it aligned with the Xpath  generated for the htmlNodes which recently been implemented in POM learning function.
An HTML document can only have one html tag If you just put several HTML document together, it will be an invalid document. so the recursive function will have to end with single html tag. cheers..

PR #567, opened 2019-01-24
Fixing parallel calls to GetAllItemsCache issue

EliranBarDavid:


Ginger/GingerCoreCommon/Repository/SolutionRepositoryItemInfo.cs
```diff
@@ -43,16 +43,19 @@ public override RepositoryFolderBase ItemRootRepositoryFolder
 
         public ObservableList<T> GetAllItemsCache()
         {
-
-            if (AllItemsCache == null)
+            string mLock = string.Empty;
```
yaronweiss: If the mLock is inside the method, how will it prevent reentry? , it should be lock on the class level, ping me if needed let's dicuss

yaronweiss: try Lock(this) + no need for mLock = "Released";

PR #572, opened 2019-01-25
Bugfix/srii multi thread

yaronweiss:


Ginger/GingerCoreCommon/Repository/SolutionRepositoryItemInfo.cs
```diff
@@ -164,5 +195,27 @@ RepositoryFolder<T> GetRepositoryFolderParent(RepositoryFolderBase repositoryFol
 
             return null;
         }
+
+        internal void AllItemsCacheRemove(T item)
+        {
+            mAllItemsCache.Remove(item);
```
MeniKadosh1: don't we need to Lock also the Remove operation so 2 Threads won't try to delete the same item and one of them will fail? 
i think we need to add Lock and before Remove check if item still exist on List

Ginger/Ginger/App.xaml.cs
```diff
@@ -770,7 +770,7 @@ public static bool SetSolution(string SolutionFolder)
 
                 //Cleanup
                 SolutionCleanup();
-
+                // !!!!!!!!!!!!! '\' is not good
                 if (!SolutionFolder.EndsWith(@"\")) SolutionFolder += @"\";
```
MeniKadosh1: i think we can remove this "\" adding If, because we now using Path.Combine and that will add missing "\" if needed

Ginger/GingerCoreCommon/Repository/ApplicationModelLib/ApplicationModelBase.cs
```diff
@@ -234,6 +234,8 @@ public enum eModelParts
             Parameters
         }
 
+
+        // !!!!!!!!!!!!!!!!!!!!!!!!!
         public static List<string> GetSoapSecurityHeaderContent(ref string txtBoxBodyContent)
```
MeniKadosh1: this was added as quick solution to support Security headers, we adding the headers as string to body which in it we combine VE valus (for tokens and other)

PR #585, opened 2019-01-28
Added the code to handle the exception

shahanemahesh:


Ginger/Ginger/Actions/UCValueExpression.xaml.cs
```diff
@@ -132,6 +132,7 @@ private void OpenExpressionEditorButton_Click(object sender, RoutedEventArgs e)
         {
             ValueExpressionEditorPage w = new ValueExpressionEditorPage(obj, AttrName, HideVEBusinessFlowAndActivityVariables);
             w.ShowAsWindow();
+            ValueTextBox.Text = Convert.ToString(obj.GetType().GetProperty(AttrName).GetValue(obj));
```
MeniKadosh1: why we still have this ValueTextBox.Text = here? why it is needed?

shahanemahesh: Whatever we select it is not updating the text on action page, there NotifyPropertyChange is not implemented so it is not updating the action page, on the other places where the "ValueExpressionEditorPage" page is called the same implementation is done, so please let me know if i should implement the "NotifyPropertyChange" this will take some time since it needs to be handled through reflection.

azure-pipelines.yml
```diff
@@ -19,7 +19,10 @@ jobs:
   - task: NuGetCommand@2
     inputs:
       restoreSolution: '$(solution)'  
+      
+  - script: dotnet restore -D:\a\1\s\Ginger    
   - task: VSBuild@1
+
```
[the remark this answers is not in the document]
shahanemahesh: i have changed the base of the branch to master so this changes were merged so the changes came into the branch, discussed with ambar these changes are required

MeniKadosh1:
this branch was approved but now have conflicts- please merge with latest and update branch for checks to be passed, we need to push it already

PR #595, opened 2019-01-31
Validate connect details exist

uzanh:


Ginger/GingerCore/ALM/QC/QCConnect.cs
```diff
@@ -102,6 +102,10 @@ public static string CurrentProject
         // Opening QC Connection 
         public static bool ConnectQCServer(string QCServerUrl, string QCUserName, string QCPassword)
         {
+            if (string.IsNullOrEmpty(QCServerUrl) || string.IsNullOrEmpty(QCUserName) || string.IsNullOrEmpty(QCPassword))
```
MeniKadosh1: usually you can't be sure that Password won't be empty but let's assume...

PR #597, opened 2019-02-01
filtering specific user process

ravirk91:
WindowDriver Filter user specific process


Ginger/GingerCore/Actions/UIAutomation/UIAComWrapperHelper.cs
```diff
@@ -434,6 +434,12 @@ public override List<AppWindow> GetListOfDriverAppWindows()
                         {
                             continue;
                         }
+
+                        Process isLocalProcess = Process.GetProcesses().FirstOrDefault(p => p.StartInfo.Environment["USERNAME"] == Environment.UserName);
```
JinendraGhodke: Why we are iterating all the processes again. We just need to if process with id window.Current.ProcessId belongs to current user or not. Like we did for java

Ginger/GingerCore/Actions/UIAutomation/UIAComWrapperHelper.cs
```diff
@@ -476,6 +481,10 @@ public override List<AppWindow> GetListOfDriverAppWindows()
 
                     foreach (AutomationElement window in AppWindows)
                     {
+                        if (CheckUserSpecificProcess(window) == false)
```
JinendraGhodke: Add this check after IsWindowValid check. Becuase sometimes windows is not valid and trying to retrieve property will throw exceptin

ravirk91: Pushed above changes

PR #600, opened 2019-02-01
Removed Invalid File Name Chars Added Generic Function to Remove Invalid Chars From File Name String

arvindgh:


Ginger/GingerCoreCommon/GeneralLib/General.cs
```diff
@@ -224,5 +224,15 @@ public static void ClearDirectoryContent(string DirPath)
             }
         }
 
+        public static string RemoveInvalidFileNameChars(string Item)
+        {
+            string fileName = Item;
+            foreach (char invalidChar in Path.GetInvalidFileNameChars())
+            {
+                fileName = fileName.Replace(invalidChar.ToString(), "");
```
JinendraGhodke: Directly use str.Replace(oldChar, newChar) instead of doing ToString

Ginger/GingerCoreCommon/GeneralLib/General.cs
```diff
@@ -224,5 +224,15 @@ public static void ClearDirectoryContent(string DirPath)
             }
         }
 
+        public static string RemoveInvalidFileNameChars(string Item)
```
JinendraGhodke: Give some valid variable name instead of item. Also use pascal casing for method parameter names

Ginger/GingerCoreCommon/GeneralLib/General.cs
```diff
@@ -224,5 +224,15 @@ public static void ClearDirectoryContent(string DirPath)
...
+        {
+            string fileName = Item;
+            foreach (char invalidChar in Path.GetInvalidFileNameChars())
+            {
+                fileName = fileName.Replace(invalidChar.ToString(), "");
+            }
+            fileName = fileName.Replace(@".", "");
```
JinendraGhodke: Are we passing the filename with extension e.g. Activity>111.xml to this method or just the name without extension.

Because if passing with extension then removing the dot will cause problem

arvindgh: We are passing only filename without extension.

PR #608, opened 2019-02-05
fix execute full report sequence issue

uzanh:


MeniKadosh1:
merged it, please update ticket and Ambar

PR #628, opened 2019-02-08
Changed sequence of test steps as per the test cases of ALM

arvindgh:
Sequence of teststep changing because of every new test group is added to after the current which is first step


Ginger/Ginger/ALM/QC/TreeViewItems/QCTestSetTreeItem.cs
```diff
@@ -96,7 +96,8 @@ public void IsTestSetAlreadyImported()
                     {
                         AlreadyImported = true;
                         MappedBusinessFlow = bf;
-                        MappedBusinessFlowPath = System.IO.Path.Combine(bf.ContainingFolder, bf.Name);                       
+                        string businessFlowName = Amdocs.Ginger.Common.GeneralLib.General.RemoveInvalidFileNameChars(bf.Name);
```
JinendraGhodke: why we removing invalid chars from here ? i think its already removed when saving file.

PR #646, opened 2019-02-13
Switch window: Adding Default Sync time 30 secs

ravirk91:
Bug 6997:Switch window action taking default sync time "-1"

### DIFF
--- a/Ginger/GingerCore/Actions/ActSwitchWindow.cs
+++ b/Ginger/GingerCore/Actions/ActSwitchWindow.cs
@@ -79,7 +79,6 @@ namespace GingerCore.Actions
             }
         }
 
-        [IsSerializedForLocalRepository]
         public Int32 WaitTime
         {
             get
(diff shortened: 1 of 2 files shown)


Ginger/GingerCore/Actions/ActSwitchWindow.cs
```diff
@@ -86,10 +86,24 @@ public Int32 WaitTime
             {
                 string val = GetInputParamValue("WaitTime");
                 Int32 intVal = -1;
+
                 if (String.IsNullOrEmpty(val))
+                {
```
JinendraGhodke: i think we can simplify it like below instead of too many if conditions. Test it by loading old xml to ensure we are not changing the value.
Please add also 2 unit test if value is not null or empty then wait time should be equal to the value specified
If value is  null or empty then it should be equal to 30

 Int32 intVal = 30;
if (!String.IsNullOrEmpty(val))
{
   Int32.TryParse(val, out intVal);
}

ravirk91: Pushed discussed changes and also added unit test

PR #654, opened 2019-02-15
Bug fix/SlowLearning with corrections

EliranBarDavid:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -3859,20 +3882,67 @@ private ElementInfo GetElementInfoWithIWebElement(IWebElement el, HtmlNode elNod
...
+                EI.Locators = ((IWindowExplorer)this).GetElementLocators(EI); 
+                ((IWindowExplorer)this).UpdateElementInfoFields(EI); 
+                EI.Properties = ((IWindowExplorer)this).GetElementProperties(EI); 
             }
 
             return EI;
         }
 
-        string GetBestElementName(ElementInfo EI)
+        string GetElementName(HTMLElementInfo EI)
```
MeniKadosh1: you modified the original "GetBestElementName" function which now based on "HTMLElementObject" just need to make sure other calls won;'t get to this function with HTMLElementObject null

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -4381,9 +4440,62 @@ ObservableList<ControlProperty> IWindowExplorer.GetElementProperties(ElementInfo
                 ElementInfo.ElementObject = el;
             }
 
-            //Learn more properties
+            //Base properties
```
MeniKadosh1: else code in row 4437 should call to function which will get the ElementObject instead implementing it in this function (getting the Xpath...).
the all idea is to stop doing the same thing over and over again in multiple functions.
this function should handle only Properties

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -4381,9 +4440,62 @@ ObservableList<ControlProperty> IWindowExplorer.GetElementProperties(ElementInfo
...
+
+                        }
+
+                    }
+
+                    if (ElementInfo.OptionalValuesObjectsList.Count > 0)
+                    {
+                        ElementInfo.OptionalValuesObjectsList[0].IsDefault = true;
+                    }
+                    list.Add(new ControlProperty() { Name = "Optional Values", Value = ElementInfo.OptionalValuesObjectsListAsString.Replace("*", "") });
```
MeniKadosh1: please enter the "list.add.." into the if, it should be added only if list is bigger than zero.
this is also the conflict you have with latest after i pushed POM Delta changes

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -3676,17 +3674,32 @@ private ObservableList<ElementInfo> GetAllElementsFromPage(string path, List<eEl
                         }
 
                         //filter element if needed
-                        eElementType foundElementType = GetElementTypeEnum(el);
+                        HTMLElementInfo foundElemntInfo = new HTMLElementInfo();
```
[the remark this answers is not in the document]
EliranBarDavid: Corrected

Ginger/GingerCore/Drivers/SeleniumDriver.cs
EliranBarDavid: do you mean to find the element by its locators?
what if it comes to this function with no locators?

Ginger/GingerCore/Drivers/SeleniumDriver.cs
EliranBarDavid: corrected,
I added solution to support old functionality as well

Ginger/GingerCore/Drivers/SeleniumDriver.cs
EliranBarDavid: Corrected

Ginger/GingerCore/Drivers/SeleniumDriver.cs
MeniKadosh1: no, just to pull the few code to seperate function, you can do it later with the massive cleanup needed

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -3687,17 +3685,35 @@ private ObservableList<ElementInfo> GetAllElementsFromPage(string path, List<eEl
...
-                            if (!filteredElementType.Contains(foundElementType))
+                            if (!filteredElementType.Contains(elementTypeEnum))
                             {
                                 continue;
                             }
                         }
 
-                        ElementInfo foundElemntInfo = null;
-                        foundElemntInfo = GetElementInfoWithIWebElement(el, htmlNode, path, learnFullElementInfoDetails);
+                        HTMLElementInfo foundElemntInfo = new HTMLElementInfo();
```
MeniKadosh1: foundElemntInfo creation needs to be inside  if (learnPartialElementInfoDetails) because on else it will create new elementInfo object in old function

PR #674, opened 2019-02-21
Couchbase DB-Not able to execute query

arvindgh:


Ginger/GingerCore/NoSqlBase/GingerCouchbase.cs
```diff
@@ -103,18 +101,40 @@ private void Disconnect()
...
+            }
+        }
+
+
+        private string GetBucketName(string inputSQL)
+        {
+            string bucketName = string.Empty;
+            if (Action== ActDBValidation.eDBValidationType.RecordCount)
+            {
+                bucketName=inputSQL.Trim();
```
JinendraGhodke: Need to replace single quotes?

arvindgh: Added and pushed

arvindgh:
> As discussed, please add exception handling i.e. if DB is not available or connection fails for any reason then action should fail

Already added, it is in calling function

PR #685, opened 2019-02-25
RemovingExceptionsFromLearnElementInfoDetailsImplementations

EliranBarDavid:


EliranBarDavid:
> how this LearnElementInfoDetails() impacted Java/Windows drivers?
> why when they do Spy it been triggered? they need to keep learn the SPy element as before, is it been done as before?

Because SelenumDriver GetElementByMouseLocator is not learning all data so we need to call the learn function, in case we are running from Java driver the Learn function will do nothing and all the data will be leared while calling the GetElementByMouseLocator.

PR #710, opened 2019-03-07
Added code to handle select all option

shahanemahesh:


Ginger/Ginger/Actions/ActionConversion/SelectActionWzardPage.xaml.cs
```diff
@@ -101,10 +101,26 @@ private void SetGridView()
...
             xGridConvertibleActions.SetTitleLightStyle = true;
+            xGridConvertibleActions.btnMarkAll.Visibility = System.Windows.Visibility.Visible;
+            xGridConvertibleActions.MarkUnMarkAllActive += MarkUnMarkAllActions;
             xGridConvertibleActions.ValidationRules = new List<ucGrid.eUcGridValidationRules>()
             {
                 ucGrid.eUcGridValidationRules.CheckedRowCount
             };
         }
+
+        private void MarkUnMarkAllActions(bool ActiveStatus)
```
MeniKadosh1: 1- you do not need all of this IF checks
2- you do not need to pull the list from the Grid datasource you already know what is the list- it's mWizard.ActionToBeConverted
3- you do not need to re-enter the list again to the Grid.DataSource because you only changing some fields in it existing objects list, you don't changing the list

please replace all this code with below simple three rows
  foreach (ConvertableActionDetails act in mWizard.ActionToBeConverted)
                {
                    act.Selected = ActiveStatus;
                }

Ginger/Ginger/Actions/ActionConversion/SelectActivityWzardPage.xaml.cs
```diff
@@ -63,6 +63,7 @@ private void SetGridsView()
             xGrdGroups.SetAllColumnsDefaultView(defView);
             xGrdGroups.InitViewItems();
             xGrdGroups.SetTitleLightStyle = true;
+            xGrdGroups.btnMarkAll.Visibility = System.Windows.Visibility.Visible;
```
MeniKadosh1: please update MarkUnMarkAllActivities function code to fit my comments in the other page

Ginger/Ginger/Actions/ActionConversion/SelectActionWzardPage.xaml.cs
shahanemahesh: Changed as suggested, but if condition will prevent the error if the list is null

MeniKadosh1:
merged, please update ticket

PR #712, opened 2019-03-07
Bugfix/bug 7160 opening of converted actions

shahanemahesh:


Ginger/Ginger/Actions/ActionConversion/ConversionConfigurationWzardPage.xaml
```diff
@@ -43,7 +43,7 @@
                 <ColumnDefinition Width="170" />
                 <ColumnDefinition Width="*" />
             </Grid.ColumnDefinitions>
-            <CheckBox x:Name="xChkDefaultTargetApp" Height="20" Grid.Column="0" Grid.Row="0"  Margin="25,5,0,0" HorizontalAlignment="Left" IsChecked="{Binding Path=DefaultTaretApp, Mode=TwoWay}">
+            <CheckBox x:Name="xChkDefaultTargetApp" Height="20" Grid.Column="0" Grid.Row="0"  Margin="25,5,0,0" HorizontalAlignment="Left" IsChecked="{Binding Path=DefaultTargetAppChecked, Mode=TwoWay}">
                 <CheckBox.Content>
```
MeniKadosh1: this is why we usually do not do Binding in XAML because in compliation time it is not catching that this field is not exist- please do binding on cs side in future,

PR #714, opened 2019-03-07
Added Prefix Before File Path to create access long file path

arvindgh:


Ginger/GingerCore/Repository/UpgradeLib/SolutionUpgrade.cs
```diff
@@ -46,7 +46,8 @@ public enum eGingerVersionComparisonResult
         {
             ConcurrentBag<Tuple<eGingerVersionComparisonResult, string>> solutionFilesWithVersion = new ConcurrentBag<Tuple<eGingerVersionComparisonResult, string>>();
             // read all XMLs and check for version
-            Parallel.ForEach(solutionFiles, FileName =>
+            //Parallel.ForEach(solutionFiles, FileName =>
+            foreach(string FileName in solutionFiles)
             {
```
MeniKadosh1: why Parallel was removed???? 
it will increase Solution loading time

Ginger/GingerCoreCommon/Repository/NewRepositorySerializer.cs
```diff
@@ -72,6 +72,9 @@ public static object OnNewRepositorySerializerEvent(NewRepositorySerilizerEventA
         public  void SaveToFile(RepositoryItemBase ri, string FileName)
         {
             string txt = SerializeToString(ri);
+
+            string LongPathPrefix = @"\\?\";
```
MeniKadosh1: 1- instead mentioning this LongPathPrefix in so many places, please add function so SolutionRepository called AdjustLongFilePath(string path) which will do the adjustment genericly and call it where needed

2- when declaring internal field it should start with lower letter like "longPathPrefix" and not "LongPathPrefix"

PR #716, opened 2019-03-07
Fixing BDD tags exception + Window explorer issue

EliranBarDavid:
Fixing WIndow Explorere issues

### DIFF
--- a/Ginger/Ginger/WindowExplorer/Common/WindowExplorerPage.xaml.cs
+++ b/Ginger/Ginger/WindowExplorer/Common/WindowExplorerPage.xaml.cs
@@ -437,9 +437,9 @@ namespace Ginger.WindowExplorer
                 StatusTextBlock.Text = "Spying Element, Please Wait...";
                 GingerCore.General.DoEvents();
                 mSpyElement = mWindowExplorerDriver.GetControlFromMousePosition();
-                mWindowExplorerDriver.LearnElementInfoDetails(mSpyElement);
                 if (mSpyElement != null)
                 {
+                    mWindowExplorerDriver.LearnElementInfoDetails(mSpyElement);
                     StatusTextBlock.Text = mSpyElement.XPath;
                     if (mSyncControlsViewWithLiveSpy)
                     {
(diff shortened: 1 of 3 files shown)


Ginger/Ginger/WindowExplorer/Common/WindowExplorerPage.xaml.cs
```diff
@@ -839,20 +839,25 @@ private void SetControlsGridView()
...
-        private void RefreshControlsGrid()
+        private async void RefreshControlsGrid()
         {
-            if (WindowsComboBox.SelectedValue != null && mWindowExplorerDriver != null)
+            if (mWindowExplorerDriver != null)
             {
-                List<ElementInfo> list = mWindowExplorerDriver.GetVisibleControls(CheckedFilteringCreteriaList.Select(x => x.ElementType).ToList());
-                                
+                List<ElementInfo> list = await Task.Run(() => mWindowExplorerDriver.GetVisibleControls(CheckedFilteringCreteriaList.Select(x => x.ElementType).ToList()));
                 // Convert to obserable for the grid
```
MeniKadosh1: why the check for "WindowsComboBox.SelectedValue != null " was removed?

EliranBarDavid: Because its impossible to interact with a UI element in async function

MeniKadosh1: but it was added from some reason so we can't just avoid and remove this check.
so do it by storing the WindowsComboBox.SelectedValue on global field and update it on WindowsComboBox.SelectionChanged event and then check global field is null in IF

EliranBarDavid: Corrected

EliranBarDavid:
> Code looks good but did you tested it for Java and Windows (grid/tree/LiveSpy)?

yes, It tested.

PR #717, opened 2019-03-07
fix config path don't set

uzanh:


MeniKadosh1:
merged , please update ticket.

PR #755, opened 2019-03-14
Fixed Test Connection for DB and password decryption

arvindgh:


Ginger/GingerCore/Environments/Database.cs
```diff
@@ -413,10 +413,11 @@ public Boolean Connect(bool displayErrorPopup = false)
                         bool isConnectionCB;
                         isConnectionCB = CouchbaseDriver.Connect();
                         if (isConnectionCB == true)
-                        { 
+                        {
                             LastConnectionUsedTime = DateTime.Now;
+                            return true;
                         }
-                        return true;
+                        else { return false; }
```
JinendraGhodke: need to add same handling also for cassandra

PR #758, opened 2019-03-14
Bug fix/7131 tefperu mail report attachments forced tobe zipped

uzanh:


Ginger/Ginger/Run/RunSetActions/RunSetActionHTMLReportSendEmailEditPage.xaml.cs
```diff
@@ -272,5 +272,15 @@ private void xcbConfigureCredential_Unchecked(object sender, RoutedEventArgs e)
             xLabelPass.Visibility = Visibility.Collapsed;
             xLabelUser.Visibility = Visibility.Collapsed;
         }
+
+        private void GridParamCheckBox_Click(object sender, RoutedEventArgs e)
+        {
```
MeniKadosh1: why is that needed and binding do not set the value inot field automatically.
i see IsChecked="{Binding ZipIt} exist on the template so better checking what is wrong with the binding than adding this click event handling

MeniKadosh1:
merged, please update ticket.

PR #771, opened 2019-03-18
Add Sub folder- fix to add it to cache

JinendraGhodke:


Ginger/GingerCoreCommon/Repository/RepositoryFolder.cs
```diff
@@ -620,7 +620,7 @@ public override RepositoryFolderBase AddSubFolder(string folderName)
             {
                 GetSubFolders();//calling it so it will find exisitng folders before adding new one
                 Directory.CreateDirectory(PathHelper.GetLongPath(FullPath));
-                GetSubFolders().Add(subfolder);
+                mSubFoldersCache?.Add(subfolder);
```
MeniKadosh1: we should not use "?" in code, it just not readable...
why we check if it is null? in which case it can be null? and if it is null so we won't add the folder?
also i was not able to reproduce the bug by following the ticket steps...

JinendraGhodke: ? Is newly added by c# and it's simplifies the code to have null check. Instead of writing full null check with if and curly braces. I thing it is more readable we just need to get used to it.

In this case i agree null check is not needed. I just added it as additional check

Issue is not reproducible on existing folders. If you add new sub folder and try to move item to it it won't work. Because newly added folder is not found in cache. If reloading solution it will work

MeniKadosh1:
I added new folders and was not getting the issue.
What was the issue? Because I don’t get what the code change is doing differently.


Regards,

Meni Kadosh
Automation Tools Development Team Lead
Amdocs Quality Engineering Services
BEAT R&D Center
+972-8-62-40961 (office)

[amdocs-2017-brand-mark-rgb]
Follow us on Facebook<http://www.facebook.com/amdocs/>, Twitter<http://twitter.com/amdocs>, LinkedIn<http://www.linkedin.com/company/amdocs>, YouTube<http://www.youtube.com/amdocsinc>, Google+<https://plus.google.com/105657940751678445194> and the Amdocs blog network<http://blogs.amdocs.com/>.

PR #775, opened 2019-03-19
Added value expression editor in runset report path

arvindgh:


Ginger/GingerCoreNET/Run/RunSetActions/RunSetActionHTMLReport.cs
```diff
@@ -51,9 +52,32 @@ public override bool SupportRunOnConfig
...
+            {
+                return mHTMLReportFolderName;
+            }
+            set
+            {
+                mHTMLReportFolderName = value;
+                OnPropertyChanged(nameof(HTMLReportFolderName));
+            }
+        }
+        ValueExpression mValueExpression = null;
```
JinendraGhodke: instead of value expression, create a field HTMLFolderNameCalculateda and use it

PR #800, opened 2019-04-01
Bug fix/7340 duplicating existing api model does not copy its model globlal parameter git bug#787

vyasprateek:
Handled GitBug#787/TFS#7340 : Duplicate the existing api model does not copy the added Global Parameters in Model

AppAPIModelTreeItem : Overriding the API Model specific Duplication mechanism 
AppAPIModelFolderTreeItem : Overriding the API Model specific Copy-Paste mechanism


Ginger/Ginger/SolutionWindows/TreeViewItems/ApplicationModelsTreeItems/AppApiModelTreeItem.cs
```diff
@@ -91,5 +94,33 @@ public void SetTools(ITreeView TV)
             
             AddSourceControlOptions(mContextMenu);
         }
+
+        public override void DuplicateTreeItem(object item)
+        {
+            if (item is RepositoryItemBase)
```
MeniKadosh1: No need the IF-Else- because for sure it is a RepositoryItemBase- it's an API model

Ginger/Ginger/SolutionWindows/TreeViewItems/ApplicationModelsTreeItems/AppApiModelsFolderTreeItem.cs
```diff
@@ -171,5 +172,38 @@ public void AddAPIModelFromDocument(object sender, RoutedEventArgs e)
         {            
             WizardWindow.ShowWizard(new AddAPIModelWizard(mAPIModelFolder), 1000);
         }
+
+        public override bool PasteCopiedTreeItem(object nodeItemToCopy, TreeViewItemGenericBase targetFolderNode, bool toRefreshFolder = true)
+        {
+            if (nodeItemToCopy is RepositoryItemBase)
+            {
```
MeniKadosh1: instead repeating the same code twice, create STATIC HandleAPIItemDuplication(APIModel source) method which returns the duplicated API model on the TreeItem class and then call it also from FolderTree Item for Paste- so then we will need to maintain only one code

PR #816, opened 2019-04-05
Exception on Ginger Close

ambarkhan:


Ginger/GingerCore/GeneralLib/GenericWindow.xaml.cs
```diff
@@ -50,12 +50,14 @@ public partial class GenericWindow : Window
...
         double mPageOriginalWidth = -1;
         double mPageOriginalHeight = -1;
+        private bool OwnerWindowClosing = false;
 
         public GenericWindow(Window Owner, eWindowShowStyle windowStyle, string windowTitle,
                                 Page windowPage, ObservableList<Button> windowBtnsList = null, bool showClosebtn = true, string closeBtnText = "Close", RoutedEventHandler closeEventHandler = null)
         {
             InitializeComponent();   
             this.Owner = Owner;
+            this.Owner.Closing += Owner_Closing;
```
MeniKadosh1: need to add Owner not null check here becuase in some cases it comes null

PR #822, opened 2019-04-09
Fix for delete an activity

aditijagtap:


Ginger/Ginger/UserControlsLib/ucGridView/ucGrid.xaml.cs
```diff
@@ -724,10 +724,14 @@ private void btnDelete_Click(object sender, RoutedEventArgs e)
 
             List<object> SelectedItemsList = grdMain.SelectedItems.Cast<object>().ToList();
 
+            Mouse.OverrideCursor = Cursors.Wait;
```
JinendraGhodke: add wait cursor before SaveUndoData

Ginger/Ginger/UserControlsLib/ucGridView/ucGrid.xaml.cs
```diff
@@ -724,10 +724,14 @@ private void btnDelete_Click(object sender, RoutedEventArgs e)
...
+
             foreach (object o in SelectedItemsList)
             {
                 mObjList.Remove(o);
             }
+
+            Mouse.OverrideCursor = null;
```
JinendraGhodke: add try catch and out it in finally

PR #823, opened 2019-04-10
FIXED : 7296-POM Wizard Multiple Agents Issue

vyasprateek:
FIXED Bug #7296-POM Wizard Multiple Agents Issue

### DIFF
--- a/Ginger/Ginger/ApplicationModelsLib/POMModels/PomElementsPage.xaml.cs
+++ b/Ginger/Ginger/ApplicationModelsLib/POMModels/PomElementsPage.xaml.cs
@@ -234,6 +234,10 @@ namespace Ginger.ApplicationModelsLib.POMModels
         internal void SetAgent(Agent agent)
         {
             mAgent = agent;
+            foreach (ElementInfo elemInfo in mElements)
+            {
+                elemInfo.ElementObject = null;
+            }
         }
 
         public enum eGridView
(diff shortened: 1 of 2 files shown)


Ginger/Ginger/ApplicationModelsLib/POMModels/PomElementsPage.xaml.cs
```diff
@@ -234,6 +234,10 @@ private void RemoveElementsToMappedBtnClicked(object sender, RoutedEventArgs e)
         internal void SetAgent(Agent agent)
         {
             mAgent = agent;
+            foreach(ElementInfo elemInfo in mElements)
+            {
+                elemInfo.ElementObject = null;
+            }
         }
```
MeniKadosh1: remove the object check IF in SeleniumDriver HighlightElement() to make sure we covering also scenario in which user changes only the page

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -4417,26 +4417,21 @@ private void HighlightElement(ElementInfo ElementInfo, bool locateElementByItLoc
                 SwitchFrame(ElementInfo.Path, ElementInfo.XPath, true);
 
                 //Find element 
-
-                if ((IWebElement)ElementInfo.ElementObject == null)
+                if (locateElementByItLocators)
```
MeniKadosh1: this is fine but why you removed the Object cleanup on POM page when Agent is been changed? please add it back on top of this change.

PR #825, opened 2019-04-11
Bugfix/bug 7301 import parameter wizard error ms acess file

shahanemahesh:


MeniKadosh1:
merged and deleted branch, please update ticket.

PR #859, opened 2019-05-06
Feature/cli3

KashmiraDeshpande:


Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -143,17 +144,19 @@ public async Task AnalyzeWithoutUI()
...
+            // Each analyzer will set to true once completed, this is prep for multi run in threads for speed
+            BusyInProcess = true;
+            mAnalyzerCompleted = false;
+            mAnalyzeDoneOnce = true;
+            try
+            {
+                if (mAnalyzeWithUI)
+                {
+                    SetStatus("Analyzing Started");
+                }
```
MeniKadosh1: why change need to be on the AnalyzerPage? what if we do not show UI, we still using the Page for running Analyzer?

Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -269,37 +276,45 @@ private List<string> RunBusinessFlowAnalyzer(BusinessFlow businessFlow, bool mar
...
+           
+            // Removing Parallel.Foreach as its freezing on main thread when running from CLI
 
-                List<string> activityVarList = AnalyzeActivity.GetUsedVariableFromActivity(activity);
-                usedVariablesInActivity.AddRange(activityVarList);
-                ReportUnusedVariables(activity, usedVariablesInActivity);
-                usedVariablesInBF.AddRange(usedVariablesInActivity);
-                usedVariablesInActivity.Clear();
-            });
+                //Parallel.ForEach(businessFlow.Activities, new ParallelOptions { MaxDegreeOfParallelism = 5 }, activity =>
```
MeniKadosh1: why Parallel analyzing was removed? this will damage the Analyzer performance

Ginger/Ginger/BusinessFlowsLibNew/NewAutomatePage.xaml.cs
```diff
@@ -73,19 +73,13 @@ public NewAutomatePage(BusinessFlow businessFlow)
 
             //WorkSpace.Instance.GingerRunner.CurrentBusinessFlow.PropertyChanged += CurrentBusinessFlow_PropertyChanged;
             //WorkSpace.Instance.GingerRunner.GingerRunnerEvent += GingerRunner_GingerRunnerEvent;
-
-
-            App.PropertyChanged += App_PropertyChanged;
+
```
MeniKadosh1: please undo newAutomatePage changes, it is not relevent and will cause later conflict because i'm working on this page

Ginger/Ginger/GeneralWindows/Run Set Pages/NewRunSetPage.xaml
```diff
@@ -41,19 +41,20 @@
...
                 </Grid.RowDefinitions>
                 <Grid.ColumnDefinitions>
-                    <ColumnDefinition Width="200*"></ColumnDefinition>
-                    <ColumnDefinition Width="270"></ColumnDefinition>
+                    <ColumnDefinition Width="8*"/>
+                    <ColumnDefinition Width="23*"/>
+                    <ColumnDefinition Width="420"></ColumnDefinition>
                 </Grid.ColumnDefinitions>
-                <StackPanel Grid.Row="0" Grid.Column="0" Orientation="Horizontal" HorizontalAlignment="Left" VerticalAlignment="Center" Margin="0,10,0,0" >
+                <StackPanel Grid.Row="0" Grid.Column="0" Orientation="Horizontal" HorizontalAlignment="Left" VerticalAlignment="Center" Margin="0,15,0,5" Height="40" Width="400" >
```
MeniKadosh1: specifiing specific width is not seems like correct solution, width should be dynamic based on screen size so for example if big screen been used it should allow showing RunSet name which is longer than 247

Ginger/Ginger/GeneralWindows/Run Set Pages/NewRunSetPage.xaml.cs
```diff
@@ -1252,52 +1262,9 @@ private void UpdateRunnersTabHeader()
...
-            {
-                foreach (char value in invalidChars)
-                {
-                    if(mRunSetConfig.Name.Contains(value))
-                        mRunSetConfig.Name = mRunSetConfig.Name.Replace(value, '_');
-                }
-            }
-
-            CreateRunSetShortCut(mRunSetConfig.Name, xRunsetEnvironmentCombo.Text);
+
```
MeniKadosh1: are we doing Run set name invalid chars check in new wizard?
foreach (char value in invalidChars)
                {
                    if(mRunSetConfig.Name.Contains(value))
                        mRunSetConfig.Name = mRunSetConfig.Name.Replace(value, '_');
                }

PR #879, opened 2019-05-17
Added condition to handle the issue

shahanemahesh:


shahanemahesh:
The issue is not able to produce, so tried to reduce the probability of issue

PR #880, opened 2019-05-17
Bug fix/7460 chrome driver launch issue using ginger

ravirk91:
Bug fix/7460 chrome driver launch issue using ginger.


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -401,10 +401,13 @@ public override void StartDriver()
                         if (SeleniumUserArgs != null)
                             foreach (string arg in SeleniumUserArgs)
                                 options.AddArgument(arg);
-                        if (HideConsoleWindow)
+                        if (HideConsoleWindow || HttpServerTimeOut != 60)
```
JinendraGhodke: why we checking if servertimeout is not 60 ?

ravirk91: I have pushed changes for above. Please review.

PR #910, opened 2019-05-24
Bugfix/bug 7491 variable gets cleared data source action

shahanemahesh:


shahanemahesh:
Mistakenly added the code for "Bug:7550 - Data source - mark all as inactive not working"

PR #912, opened 2019-05-24
Fixed UpdateMany for MongoDB

arvindgh:


Ginger/GingerCore/NoSqlBase/GingerMongoDb.cs
```diff
@@ -292,12 +292,26 @@ public override void PerformDBAction()
                         {
                             var session = mMongoClient.StartSession();
                             session.StartTransaction();
-                            collection.UpdateOne(filterDocumnet, paramDocumnet);
+                            if (SQLCalculated.Contains("updateMany"))
```
JinendraGhodke: Line 295 -302 is duplicate in if as well as else. Please put in a function and call the function in both the places

PR #914, opened 2019-05-24
Bug fix/7524 shared repository global variable fix

aditijagtap:


Ginger/Ginger/Variables/VariableEditPage.xaml.cs
```diff
@@ -77,17 +77,25 @@ public VariableEditPage(VariableBase v, Context context, bool setGeneralConfigsA
             GingerCore.GeneralLib.BindingHandler.ObjFieldBinding(cbSetAsInputValue, CheckBox.IsCheckedProperty, mVariable, nameof(VariableBase.SetAsInputValue));
             GingerCore.GeneralLib.BindingHandler.ObjFieldBinding(cbSetAsOutputValue, CheckBox.IsCheckedProperty, mVariable, nameof(VariableBase.SetAsOutputValue));
 
-            if (mode ==eEditMode.Global)
+            if (mode ==eEditMode.Global || mode==eEditMode.SharedRepository)
```
JinendraGhodke: With this condition set as input and set as outpit check boxes will be hidden when opening variable from Shared repository.
But this check boxes should be hidden only for edit mode global and not for shared repository

Ginger/Ginger/Variables/VariablesPage.xaml.cs
```diff
@@ -367,8 +367,16 @@ private void VariablesGrid_grdMain_MouseDoubleClick(object sender, EventArgs e)
         {
             VariableBase selectedVarb = (VariableBase)grdVariables.CurrentItem;
             selectedVarb.NameBeforeEdit = selectedVarb.Name;
-
-            VariableEditPage w = new VariableEditPage(selectedVarb, mContext, false);
+            VariableEditPage.eEditMode editMode = VariableEditPage.eEditMode.BusinessFlow;
+            if (mContext.BusinessFlow == null)
```
JinendraGhodke: in few places we are passing context as null. So please add null check for mContext.
Otherwise we might get null exception when mContext is null

PR #917, opened 2019-05-24
Fixed MongoDB support for insert queries in FreeSql

arvindgh:


Ginger/GingerCore/NoSqlBase/GingerMongoDb.cs
```diff
@@ -264,14 +264,35 @@ public override void PerformDBAction()
...
+
+                        if (SQLCalculated.Contains("insertOne"))
+                        {
+                            BsonDocument insertDocumnet = BsonDocument.Parse(GetUpdateQueryParams(SQLCalculated));
+                            collection.InsertOne(insertDocumnet);
+                        }
+                        else if (SQLCalculated.Contains("insertMany"))
+                        {
+                            //need to find way to use insertMany()
+                            string queryParam = GetUpdateQueryParams(SQLCalculated).ToString();
```
JinendraGhodke: There is method insert many in collection. Why that can not be used instead of inserting one by one ?

PR #920, opened 2019-05-24
fix for test database connection

KashmiraDeshpande:


Ginger/Ginger/Environments/AppDataBasesPage.xaml.cs
```diff
@@ -80,7 +80,15 @@ private void TestDBConnection(object sender, RoutedEventArgs e)
         {
             try
             {
-                Database db = (Database)grdAppDbs.grdMain.CurrentItem;
+                Database db;
+                if ((Database)grdAppDbs.grdMain.CurrentItem != null)
+                {
+                     db = (Database)grdAppDbs.grdMain.CurrentItem;
+                }
+                else
```
JinendraGhodke: if not row selected why to set first row as default?
If current item is null that means now row is selected then we should give message please select item

PR #939, opened 2019-05-27
Added the code to handle the  7582 & 7584 bugs

shahanemahesh:


shahanemahesh:
Both bugs are fixed - 7582 & 7584

Ginger/Ginger/UserControlsLib/ucGridView/ucGrid.xaml.cs
```diff
@@ -344,8 +347,15 @@ private void TagsViewer_TagsStackPanlChanged(object sender, EventArgs e)
             grdMain.CancelEdit();
             this.Dispatcher.Invoke(() =>
             {
-                CollectFilterData();
-                mCollectionView.Refresh();
+                try
```
MeniKadosh1: why this try-catch? what is the exception hapenning and from where exactly?

shahanemahesh: When the businessflow is running then in when the execution tries to switch to 2nd activity it is giving exception as "The calling thread cannot access the object because different thread owns it" on the line "mCollectionView.Refresh();" form above file, this is raised only for the first activity, after that it runs smoothly. tried to handle the dispature invoke if required by checking access to code but it was still giving exception.

PR #975, opened 2019-06-11
Fixed : Update element not working as expected

vyasprateek:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -7269,7 +7269,13 @@ public ElementInfo GetMatchingElement(ElementInfo element, ObservableList<Elemen
             if (OriginalElementInfo == null)
             {
                 //try by type and Xpath comparison
-                OriginalElementInfo = existingElemnts.Where(x => (x.ElementTypeEnum == element.ElementTypeEnum) && (x.XPath == element.XPath) && (x.Path == element.Path || string.IsNullOrEmpty(x.Path) && string.IsNullOrEmpty(element.Path))).FirstOrDefault();
+                OriginalElementInfo = existingElemnts.Where(x => (x.ElementTypeEnum == element.ElementTypeEnum)
+                                                                    && (x.XPath == element.XPath)
+                                                                    && (x.Path == element.Path || (string.IsNullOrEmpty(x.Path) && string.IsNullOrEmpty(element.Path)))
+                                                                    && (x.Locators.FirstOrDefault(l => l.LocateBy == eLocateBy.ByRelXPath) != null && element.Locators.FirstOrDefault(l => l.LocateBy == eLocateBy.ByRelXPath) != null &&
```
MeniKadosh1: so that means that if existingElement won't have ByRelXPath locator so it won't be found, no?
we need that if existingElement don't have ByRelXPath so comparison will be done based on Xpath and Path only and if it has so it also consider the RelXpath

vyasprateek: check added. Now, we'll consider RelativeXPath only if the previous element have it.

MeniKadosh1:
merged, please update ticket

PR #978, opened 2019-06-12
Variable reset fix

KashmiraDeshpande:


Ginger/GingerCoreCommon/VariablesLib/VariableList.cs
```diff
@@ -50,7 +50,14 @@ public override string VariableUIType
 
         public override string GetFormula()
         {
```
MeniKadosh1: please explain why changes were needed in VariableList class

Ginger/GingerCoreCommon/VariablesLib/VariableSelectionList.cs
```diff
@@ -48,9 +48,24 @@ public string OptionalValues
 
         [IsSerializedForLocalRepository]
         public ObservableList<OptionalValue> OptionalValuesList = new ObservableList<OptionalValue>();
```
MeniKadosh1: why "SelectedValue" is needed?

Ginger/GingerCoreCommon/VariablesLib/VariableList.cs
KashmiraDeshpande: Because we are doing ResetValue() in the constructor. And the values of Formula and ValueList are coming null as the first time, as the constructor is being called before the activities are loaded. So in ValueList class there is just the condition of Null Value check to avoid the NullValueException.

Ginger/GingerCoreCommon/VariablesLib/VariableSelectionList.cs
```diff
@@ -48,9 +48,24 @@ public string OptionalValues
...
-
+        [IsSerializedForLocalRepository]
         public string SelectedValue { set { Value = value; OnPropertyChanged(nameof(SelectedValue)); } get { return Value; } }
 
+        private string mValue;
+
+        public override string Value
```
JinendraGhodke: we decided to override and serialize the value field for selection list variable. no?

If we serialize value then selected value serialize is not needed

Ginger/GingerCoreCommon/VariablesLib/VariableSelectionList.cs
```diff
@@ -82,8 +102,19 @@ private ObservableList<OptionalValue> ConvertOptionalValuesStringToList(string v
 
         public override void ResetValue()
         {
+            // As the initial value is not configured for VariableSelectionList
             if (OptionalValuesList.Count > 0)
-                Value = OptionalValuesList[0].Value;
+            {
+                if (SelectedValue != null)
```
JinendraGhodke: Do we still need this change ?

Ginger/GingerCoreNETUnitTest/SolutionTestsLib/VaribaleRepositorySerializerTest.cs
```diff
@@ -0,0 +1,154 @@
...
+        public void VariablesCount()
+        {
+            // Assert   
+            Assert.AreEqual(8, BusinessFlow.Variables.Count, "Variable Count");
+        }
+
+        [TestMethod]
+        public void VariableString()
+        {
+            foreach (VariableBase variable in varList)
```
JinendraGhodke: please split it to different unit tests. Because let's say even single variable load will fail it will fail the whole unit test.
Ideally 1 unit should test 1 functionality

Ginger/GingerCoreNETUnitTest/SolutionTestsLib/VaribaleRepositorySerializerTest.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        [ClassInitialize]
+        public static void ClassInitialize(TestContext TC)
+        {
+            string TempSolutionFolder = TestResources.GetTestTempFolder(@"Solutions"+ Path.DirectorySeparatorChar+ "Variables");
+            if (Directory.Exists(TempSolutionFolder))
+            {
+                Directory.Delete(TempSolutionFolder, true);
+            }
+            SR = GingerSolutionRepository.CreateGingerSolutionRepository();
```
JinendraGhodke: I think loading a BF from file using serializer is enough, do we need to create solution repository here ?

Ginger/GingerCoreNETUnitTest/SolutionTestsLib/VaribaleRepositorySerializerTest.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            SR.CreateRepository(TempSolutionFolder);
+            NewRepositorySerializer.AddClassesFromAssembly(typeof(BusinessFlow).Assembly);
+            SR.Open(TempSolutionFolder);
+            string FileName = TestResources.GetTestResourcesFile(@"Solutions"+ Path.DirectorySeparatorChar+ "Variables" + Path.DirectorySeparatorChar + "BusinessFlows" + Path.DirectorySeparatorChar + "Business Flow 1.Ginger.BusinessFlow.xml");
```
JinendraGhodke: Please change the Business flow xml and flow name to Variable Testing Flow or something relevant

PR #981, opened 2019-06-14
implementing feature for browser logs

ravirk91:
Bug 7623:Feature for preserving the network logs in chrome browser


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -5969,6 +5969,31 @@ public void ActBrowserElementHandler(ActBrowserElement act)
                     break;
                 case ActBrowserElement.eControlAction.CloseAll:
                     Driver.Quit();
+                    break;
+                case ActBrowserElement.eControlAction.GetBrowserLog:
+
```
JinendraGhodke: Please add unit test to cover this.
Also send me the screenshot of how it looks in the output values

ravirk91: added unit test

PR #990, opened 2019-06-18
Feature/lite db support share folder functionality

uzanh:


Ginger/GingerCoreNET/Run/RunSetActions/RunSetActionHTMLReport.cs
```diff
@@ -151,6 +160,45 @@ public override void Execute(ReportInfo RI)
             }
         }
 
+        private void ProduceLiteDBReportFolder(HTMLReportsConfiguration currentConf)
```
MeniKadosh1: don't we supposed to place this method in LiteDB report related class? why is it in the run set opearion class? what about mail report operation, it won't need that as well?

uzanh: 1. don't think related to operation.
2. RunSetActionHTMLReport.cs - object in the "Produce HTML Reprot" operation( inheritance from "RunSetActionBase" )
The object is in RunSetActionHTMLReportEditPage.xaml.cs
3. About the mail report - i think he will need to use it, but need to ask liran.
but then just to remove the function to Base class

PR #1000, opened 2019-06-20
Fix for Parametres are not coming automatically after running script …

aditijagtap:
…action


Ginger/Unit Tests/NonUITests/ScriptActionTest.cs
```diff
@@ -0,0 +1,77 @@
...
+            mGR.BusinessFlows.Add(mBF);
+            Reporter.ToLog(eLogLevel.DEBUG, "Creating the GingerCoreNET WorkSpace");
+            WorkSpaceEventHandler WSEH = new WorkSpaceEventHandler();
+            WorkSpace.Init(WSEH);
+            WorkSpace.Instance.SolutionRepository = GingerSolutionRepository.CreateGingerSolutionRepository();
+        }
+
+        [TestMethod]
+        public void RunScriptAction()
+        {
```
JinendraGhodke: Please add unit test for all 3 cases
1. When RC_start index is 0
2. When index is >0
3. When index is -1

Ginger/Unit Tests/NonUITests/ScriptActionTest.cs
```diff
@@ -0,0 +1,77 @@
...
+using GingerCore.Actions;
+using GingerTestHelper;
+using GingerWPF.WorkSpaceLib;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace UnitTests.NonUITests
+{
+    [TestClass]
+    [Level3]
+    public class ScriptActionTest
```
JinendraGhodke: We don't need to create new file and class only for script action test.
Please add them in NonDriverAction tests class itself

PR #1004, opened 2019-06-20
Fixed Flow Control Action missing error

arvindgh:
If there is flow control in the action and current action is changed ActionEditPage will change according to the current action

### DIFF
--- a/Ginger/Ginger/Actions/ActionEditPage.xaml.cs
+++ b/Ginger/Ginger/Actions/ActionEditPage.xaml.cs
@@ -670,21 +670,11 @@ namespace Ginger.Actions
 
             mContext.Runner.RunInSimulationMode = originalSimulationFlagValue;
 
-            /// If there is flow control in the action and current action is changed ActionEditPage will change according to the current action
-            ///
-            this.Dispatcher.Invoke(() =>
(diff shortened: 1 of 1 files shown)


Ginger/Ginger/Actions/ActionEditPage.xaml.cs
```diff
@@ -669,6 +669,12 @@ private async void RunActionButton_Click(object sender, RoutedEventArgs e)
             int res = await RunAction().ConfigureAwait(false);
 
             mContext.Runner.RunInSimulationMode = originalSimulationFlagValue;
+
+            Act tempAct = (Act)mActParentBusinessFlow.CurrentActivity.Acts.CurrentItem;
```
JinendraGhodke: When we are executing Flow control GotoAction we already update current item to new action.
So please check why that is not updating action also for edit page.
We should  not do it here explicitly

arvindgh: If Action A have flow control to action B, so the current item will be the b after run in edit page of A, but edit page we opened should remains for A.
so we are checking if the current action changed, then changing it to the previous one as current action.

PR #1018, opened 2019-06-25
Bug fix/7408 scroll bar not working in Value Expression

aditijagtap:


Ginger/GingerCore/Actions/UIAutomation/UIAComWrapperHelper.cs
```diff
@@ -3611,6 +3611,20 @@ public override void ScrollDown(object obj)
                         winAPI.SendClickOnXYPoint(element, x1, y1);
                     }
                     break;
+                case "treeview":
+                    element.TryGetCurrentPattern(ScrollPatternIdentifiers.Pattern, out vp);
```
JinendraGhodke: What if some element do not support scroll pattern.
VP will be null. Add null check for VP. 
Also variable names to be changed as per naming conventions

PR #1023, opened 2019-06-27
Handled  bug - Actions Convertion Wizard- Shows Same Action Type multilpe times

shahanemahesh:


Ginger/Ginger/Actions/ActionConversion/SelectActionWzardPage.xaml.cs
```diff
@@ -96,7 +138,7 @@ private void SetGridView()
 
             view.GridColsView.Add(new GridColView() { Field = nameof(ConvertableActionDetails.Selected), Header = "Select", WidthWeight = 3.5, MaxWidth = 50, StyleType = GridColView.eGridColStyleType.CheckBox, BindingMode = System.Windows.Data.BindingMode.TwoWay });
             view.GridColsView.Add(new GridColView() { Field = nameof(ConvertableActionDetails.SourceActionTypeName), WidthWeight = 15, Header = "Source Action Type" });
```
MeniKadosh1: Why source Activity was removed? 
we need to show the Activities containing those Action types, you need to agragarte all Activities relevent in list like "Activity1, Activity2,..." and show it

BTW- this Source Colum should not be shown in multi BF's convertion

shahanemahesh: implemented

Ginger/GingerCoreNET/ActionsLib/ActionsConversion/ActionConversionUtils.cs
```diff
@@ -228,23 +228,36 @@ public ObservableList<ConvertableActionDetails> GetConvertableActivityActions(Li
...
+                                newConvertibleActionType.ActivityList.Add(convertibleActivity.ActivityName);
+                                lst.Add(newConvertibleActionType);
+                            }
+                            else
+                            {
+                                if (!existingConvertibleActionType.Actions.Contains(act))
+                                {
+                                    existingConvertibleActionType.ActionCount++;
+                                    existingConvertibleActionType.Actions.Add(act);
+                                    existingConvertibleActionType.ActivityList.Add(convertibleActivity.ActivityName);
```
MeniKadosh1: do you check that we won't have same Activity name twice in existingConvertibleActionType.ActivityList?

shahanemahesh: Yes checked same activity not added

PR #1039, opened 2019-07-02
Bug fix/reload panel add action pages on environment or solution change

vyasprateek:
Solution change handled


Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/MainAddActionsNavigationPage.xaml.cs
```diff
@@ -71,6 +71,21 @@ private void Context_PropertyChanged(object sender, System.ComponentModel.Proper
                     LoadActionFrame(null); 
                 }                
             }
+            else if(e.PropertyName is nameof(mContext.Environment))
```
MeniKadosh1: why to reset all each time user changes the selected environment? users doing it a lot during use of Automate page with same BF/Solution

Ginger/Ginger/BusinessFlowPages_New/NewAutomatePage.xaml.cs
```diff
@@ -972,6 +977,9 @@ private void WorkSpacePropertyChanged(object sender, System.ComponentModel.Prope
                 }
 
                 UpdateToNewSolution();
+
```
MeniKadosh1: 1- move it inside the null check IF so it will be reset as part of Solution change reset
2- branch has Conflicts- please sync it with latest and solve conflicts

Ginger/Ginger/UserControlsLib/UCListView/UcListView.xaml.cs
```diff
@@ -137,27 +138,35 @@ public IObservableList DataSourceList
                     }
 
                     mObjList = value;
-                    this.Dispatcher.Invoke(() =>
+
+                    xSearchTextBox.Text = "";
```
MeniKadosh1: why xSearchTextBox.Text = ""; was pulled out from .Dispatcher.Invoke(() this may cause issues in case diffrent thread will set the list DataSource because here we modify UI control- i will merge the code but i belive that will cause issue

PR #1049, opened 2019-07-07
Bugfix NewAutomatePage high bugs

vyasprateek:
=>Double clicking on "Browser Action" & "UIElement Action": throwing an exception

=>Adding Legacy Action do not ask if to add new action instead

=>On double click on library action we should add the action and not open edit page

=>API Models add action is shown even due we are on non Webservices platform Activity

=>Adding API Model action from libarary shows exception on edit


Ginger/Ginger/BusinessFlowPages_New/ActionsListViewPage.xaml.cs
```diff
@@ -168,8 +168,36 @@ private void listActions_ItemDropped(object sender, EventArgs e)
                 Act instance = null;
                 if (droppedItem is Act)
                 {
-                    Act a = droppedItem as Act;
-                    instance = (Act)a.CreateInstance(true);
+                    Act selectedAction = droppedItem as Act;
+                    if (selectedAction.AddActionWizardPage != null)
```
MeniKadosh1: we should not have duplicated code for adding Action into Actions list like we have here and in ActionsLibrarayNavPage, put the code in a method called "AddActionsHandler()" on some class, make it static, provide to it the actions to add and actions list to add to and it will do the rest.
so that way we will need to maintain only one place which will work for D&D and add from NavPage.
let me know if not clear

PR #1052, opened 2019-07-08
Bugfix/bug new automate page issues

shahanemahesh:


shahanemahesh:
On local it is not giving any error, but dont know why it is failing will check it again, the reference to the "AgentMappingPage" was already getting used on this page,

PR #1062, opened 2019-07-11
Fix for Switch Window by Index

aditijagtap:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -5720,6 +5720,18 @@ private void SwitchWindow(Act act)
                                     break;
                                 }
                             }
+                            if (act.LocateBy == eLocateBy.ByIndex)
+                            {
+                                int getWindowIndex = Int16.Parse(act.LocateValue);
```
JinendraGhodke: if locate value is using value expression then it will not work.
Please use locatevalueCaluted

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -5720,6 +5720,18 @@ private void SwitchWindow(Act act)
                                     break;
                                 }
                             }
+                            if (act.LocateBy == eLocateBy.ByIndex)
```
JinendraGhodke: Add unit test to cover this scenario

PR #1075, opened 2019-07-15
Bug Fixes

vyasprateek:
=> Getting Old edit Action Page when trying to add action from window explorer section over window explorer page and live spy page

=> Getting Legacy Action Warning pop up while adding window action and Generic Action for Java Driver

=> Adding POM's Frame element action doesn't maps POM and Element on adding

=> Deleting an Activity generates exception from WindowsExplorer and LiveSpy Pages

### DIFF
--- a/Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/ActionsFactory.cs
+++ b/Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/ActionsFactory.cs
@@ -23,6 +23,8 @@ namespace Ginger.BusinessFlowPages_New.AddActionMenu
 {
     class ActionsFactory
     {
+        public static bool IsLegacyTabSelected = false;
+
         public static void AddActionsHandler(object mItem, Context mContext)
         {
             Act instance = null;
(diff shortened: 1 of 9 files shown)


Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/LiveSpyNavPage.xaml.cs
```diff
@@ -102,13 +102,13 @@ private void SetFrameEnableDisable()
         /// This method is used to get the new WindowExplorerPage based on Context and Agent
         /// </summary>
         /// <returns></returns>
-        private void LoadWindowExplorerPage(Context context)
+        private void LoadWindowExplorerPage()
         {
```
MeniKadosh1: why we have 'LoadWindowExplorerPage()' method in LiveSpy page?

Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/ActionsFactory.cs
```diff
@@ -95,7 +97,7 @@ static Act GenerateSelectedAction(Act selectedAction, Context mContext)
             else
             {
                 Act instance = (Act)selectedAction.CreateCopy();
-                if (selectedAction is IObsoleteAction)
+                if (selectedAction is IObsoleteAction && IsLegacyTabSelected)
                 {
```
MeniKadosh1: please explain why IsLegacyTabSelected was added and IObsoleteAction was not enough

vyasprateek: There exist some actions which implements IObsoleteAction and aren't  listed under Legacy Actions. like WindowsAction on Java Platform implements IObsoleteAction but still is listed under Platform Actions and that's the reason Ravi reported an issue where,"LegacyAction Warning Pop Up is being generated on adding Windows Action over Java Platform"

Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/LiveSpyNavPage.xaml.cs
vyasprateek: I've just updated the method signature removing context parameter but haven't visited the code. I'll go through this once and revert.

PR #1077, opened 2019-07-16
Updated the name of object in LiveSpyPage

shahanemahesh:


Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/LiveSpyNavPage.xaml.cs
```diff
@@ -105,9 +105,9 @@ private void SetFrameEnableDisable()
         private void LoadWindowExplorerPage(Context context)
```
MeniKadosh1: Name of method "LoadWindowExplorerPage()" is still wrong, it should be "LoadLiveSpyPage()"- please change it on next branch you creating

Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/LiveSpyPage.xaml.cs
```diff
@@ -249,7 +249,7 @@ private void timenow(object sender, EventArgs e)
                 if (mSpyElement != null)
                 {
                     xWinGridUC.mWindowExplorerDriver.LearnElementInfoDetails(mSpyElement);
-                    StatusTextBlock.Text = mSpyElement.ElementName;
+                    StatusTextBlock.Text = mSpyElement.XPath;
```
MeniKadosh1: Why to show XPATH and not Element Name?

PR #1087, opened 2019-07-18
Bug fix/list view filter issue and shared repo action add

vyasprateek:
=> list view filter issue fixed
=> Adding actions from shared repository creates instance not copy, using Action's Containing Folder


Ginger/Ginger/BusinessFlowPages_New/AddActionMenu/ActionsFactory.cs
```diff
@@ -108,7 +108,17 @@ static Act GenerateSelectedAction(Act selectedAction, Context mContext)
             }
             else
             {
-                Act instance = (Act)selectedAction.CreateCopy();
+                Act instance = null;
```
MeniKadosh1: Still notice that add Action from Shared Repo action page using Arrow and not D&D is not using ActionsFactory and it should- please modify it on next PR

PR #1108, opened 2019-07-23
Added code to handle the POM creation while recording

shahanemahesh:


Ginger/GingerCoreNET/RecordingLib/RecordingManager.cs
```diff
@@ -106,9 +106,14 @@ private POMObjectRecordingHelper GetNewPOM(string pageTitle, string pageURL, str
                 newPOM.Guid = new Guid();
                 newPOM.ItemName = pageTitle;
                 newPOM.PageURL = pageURL;
+                newPOM.TargetApplicationKey = new RepositoryItemKey() { ItemName = Context.Target.ItemName, Guid = Context.Target.Guid, Key = Context.Target.Key.Key };
```
MeniKadosh1: you supposed only to do:
 newPOM.TargetApplicationKey =Context.Target.Key;

shahanemahesh: i tried that but after the POM is created when you go to "Resources-->POM" tab and check the newly created POM then it gives error that the T.App does not exist, so tried way

PR #1112, opened 2019-07-24
Re-analyze High critical count fix

KashmiraDeshpande:


Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -153,7 +153,7 @@ private async void RerunButton_Click(object sender, RoutedEventArgs e)
                 Reporter.ToUser(eUserMsgKey.StaticInfoMessage, "Please wait for current process to end.");
                 return;
             }
-
+            mIssues.CollectionChanged += MIssues_CollectionChanged;
```
JinendraGhodke: if we add the event again not sure if the event will be called twice.

When we are doing reanalyze we are on the same page, so we can check why event is not called for renalyze

PR #1118, opened 2019-07-26
Create new Datasource fix

KashmiraDeshpande:


Ginger/Ginger/DataSource/AddNewDataSourcePage.xaml.cs
```diff
@@ -210,6 +214,8 @@ private void New_Checked(object sender, RoutedEventArgs e)
             DSTypeComboBox.SelectedItem = DataSourceBase.eDSType.MSAccess;
             DSName.Text = "";
             FilePathTextBox.Text = "";
+            FilePathTextBox.IsEnabled = true;
+            DSName.IsEnabled = true;
```
JinendraGhodke: we need to enable FileBrowse button as well ?

PR #1122, opened 2019-07-29
Bug fix/7733 need to handle to add actuielement from new live spy page java driver and also table action is not populating with correct action on spying

ravirk91:
Bug fix/7733 need to handle to add actuielement from new live spy page java driver and also table action is not populating with correct action on spying


Ginger/GingerCore/Platforms/PlatformsInfo/JavaPlatform.cs
```diff
@@ -72,25 +72,70 @@ public override ObservableList<Act> GetPlatformElementActions(ElementInfo elemen
...
+                            );
+                    }
                 }
             }
+            
             return UIElementsActionsList;
         }
 
+        private List<ActUIElement> CreateActUITableAction(List<ActUIElement.eTableAction> tableActionList)
+        {
```
JinendraGhodke: What is this method ? do we need it or can be removed?

Ginger/GingerCore/Platforms/PlatformsInfo/JavaPlatform.cs
```diff
@@ -72,25 +72,70 @@ public override ObservableList<Act> GetPlatformElementActions(ElementInfo elemen
             ObservableList<Act> UIElementsActionsList = new ObservableList<Act>();
             eElementType elementType = GetElementType(elementInfo.ElementType);
 
-            var actionList = GetPlatformUIElementActionsList(elementType);
+            if (elementType.Equals(eElementType.Table))
+            {
+                //get all action list supported to tablecell action
+                var tableActionList = GetTableControlActions(ActUIElement.eElementAction.TableCellAction);
```
JinendraGhodke: instead of calling this method 3 times we can try to get it in one go using tuples like it is done in Web

Ginger/GingerCore/Platforms/PlatformsInfo/JavaPlatform.cs
```diff
@@ -72,25 +72,70 @@ public override ObservableList<Act> GetPlatformElementActions(ElementInfo elemen
...
+                    }
 
-            if (actionList.Count > 0)
+                    var actUITableAction = new ActUIElement()
+                    {
+                        Description = action + " : " + elementInfo.ElementTitle,
+                        ElementType = eElementType.Table,
+                        ElementAction = elementAction,
+                    };
+                    actUITableAction.GetOrCreateInputParam(ActUIElement.Fields.RowSelectorRadioParam, "RowNum");
```
JinendraGhodke: if element action is Table Action then we don't need row and all the details

Ginger/Ginger/WindowExplorer/Java/JavaTableTreeItem.cs
```diff
@@ -20,44 +20,34 @@
...
+using Amdocs.Ginger.Repository;
 
 namespace Ginger.WindowExplorer.Java
 {
     public class JavaTableTreeItem : JavaElementTreeItem, ITreeViewItem, IWindowExplorerTreeItem
     {
       ObservableList<Act> mAvailableActions = new ObservableList<Act>();
-      ActTableEditPage mActTableEditPage = null;
-
+      UIElementTableConfigPage mActTableEditPage = null;
```
JinendraGhodke: rename also the variable name to make it clear

Ginger/GingerCore/Platforms/PlatformsInfo/JavaPlatform.cs
ravirk91: will remove this method.

PR #1133, opened 2019-07-30
Update DS Lite- from Output values

KashmiraDeshpande:


Ginger/GingerCoreNET/ValueExpressionLib/ValueExpression.cs
```diff
@@ -714,7 +714,26 @@ public void ReplaceDataSource(string p)
                         {
                             Markasdone = true;
                         }
-                        mValueCalculated = liteDB.GetResut(litedbquery, col[0], Markasdone);
+                        if (litedbquery.Contains("update"))
+                        {
+                            if (litedbquery.Contains("where"))
+                            {
```
JinendraGhodke: Code here is very sensitive with lot of string operations. Let's discuss what's going on in the code and if we can simplify it. We will also decide if to do it now or after release.

Also need to add unit test for this

PR #1134, opened 2019-07-30
GDI+ Exception Fix

KashmiraDeshpande:


Ginger/Ginger/Repository/RepositoryItemFactory.cs
```diff
@@ -560,8 +560,10 @@ public void CreateCustomerLogo(object a, string tempFolder)
         {
             HTMLReportConfiguration currentTemplate = (HTMLReportConfiguration)a;
             System.Drawing.Image CustomerLogo = Ginger.General.Base64StringToImage(currentTemplate.LogoBase64Image.ToString());
-            CustomerLogo.Save(tempFolder + "/CustomerLogo.png");
-           
+            if (Directory.Exists(tempFolder))
```
JinendraGhodke: what is tempFolder here ? if this directory do not exist we should create it no ?

PR #1146, opened 2019-07-31
Bug fix/official release related feature requests and bugs pra teek

vyasprateek:
Bugs fixed and Features Implemented !


Ginger/Ginger/BusinessFlowPages_New/ActionsListViewPage.xaml.cs
```diff
@@ -133,12 +134,15 @@ private void SetListView()
             mActionsListView.ItemDropped += listActions_ItemDropped;
 
             mActionsListView.List.MouseDoubleClick += ActionsListView_MouseDoubleClick;
+
+            // Enable Virtualization for Actions ListView to improve the loading time/performance
```
MeniKadosh1: No need to set the "CanContentScroll" as true because it is True as defualt, only in case of False like in Acivities which should set it from code

Ginger/Ginger/BusinessFlowPages_New/VariabelsListViewPage.xaml.cs
```diff
@@ -172,6 +172,8 @@ private void SetListView()
             mVariabelsListView.ItemDropped += ListVars_ItemDropped;
 
             mVariabelsListView.List.MouseDoubleClick += VariabelsListView_MouseDoubleClick;
+
```
MeniKadosh1: Not needed bacause it is True as defualt

Ginger/Ginger/DragDropLib/DragDrop2.cs
```diff
@@ -83,11 +88,45 @@ private static void DragSource_Drop(object sender, DragEventArgs e)
...
+                if (sender is UcListView)
+                {
+                    _DroppedPoint = e.GetPosition(sender as UcListView);
+                }
                 ((IDragDrop)DragInfo.DragTarget).Drop(DragInfo);
             }
         }
 
-      
+        public static void ShuffleControlsItems(RepositoryItemBase draggedItem, RepositoryItemBase itemDroppedOver, UcListView xUCListView)
```
MeniKadosh1: please change "ShuffleControlsItems" to "MoveInternalDroppedItem"

PR #1180, opened 2019-08-06
datasource where conditions fix

KashmiraDeshpande:


Ginger/GingerCoreCommon/Repository/NewRepositorySerializer.cs
```diff
@@ -669,7 +669,16 @@ private static object xmlReadObject(Object Parent, XmlReader xdr, RepositoryItem
 
                 if (targetObj == null)
                 {
-                    obj = CreateObject(className);
+                    bool isClassToSkip = HandleClassToSkip(xdr, className);
```
yaronweiss: change to bool isHandled

Ginger/GingerCoreCommon/Repository/NewRepositorySerializer.cs
```diff
@@ -1240,7 +1248,25 @@ public static string GetXMLGingerVersion(string xml, string xmlFilePath)
             }
         }
 
+        private static bool HandleClassToSkip(XmlReader xdr, string className)
+        {
+            //TODO: If need to add more classed then create a dictionary of Classes to Skip
```
yaronweiss: Change to switch case

yaronweiss: remove comment no need for dictionary, each class can have different operation

Ginger/GingerCoreCommon/Repository/NewRepositorySerializer.cs
```diff
@@ -1240,7 +1248,25 @@ public static string GetXMLGingerVersion(string xml, string xmlFilePath)
...
+            if (className == "GingerCore.DataSource.ActDSConditon")
+            {
+                int level = xdr.Depth;
```
yaronweiss: Create a function to move xdr to next element, so in case needed can be used in other class case

PR #1237, opened 2019-08-22
Fix for XML/JSON Validation Action for xml is confusing

aditijagtap:


Ginger/GingerCoreTest/Misc/NonDriverActionTest.cs
```diff
@@ -309,7 +310,90 @@ public void RestartAgentNotRunningTest()
...
-        
 
+        [TestMethod]
+        public void XMLRequestFromFileCheckedTestForPath()
+        {
+            //Arrange
+            ActXMLTagValidation action = new ActXMLTagValidation();
+            action.DocumentType = GingerCore.Actions.XML.ActXMLTagValidation.eDocumentType.XML;
+            action.ReqisFromFile = true;
+            action.InputFile.Value = TestResources.GetTestResourcesFile(@"XML\book.xml");
```
JinendraGhodke: do not use slash in the file paths. It will not work on linux. Please use Path.DirectorySeparator

PR #1278, opened 2019-09-10
Bug fix/email run set elapsed detail is wrong

uzanh:


Ginger/Ginger/Reports/HTMLReportAttachmentConfigurationPage.xaml.cs
```diff
@@ -100,26 +100,15 @@ private void OkButton_Click(object sender, RoutedEventArgs e)
...
                 }
-                else if(extraInformationCalculated.Length > 100)
+                else if (extraInformationCalculated.Length > 100)
                 {
                     Reporter.ToUser(eUserMsgKey.FolderNamesAreTooLong);
                     return;
-                }                
-                else
+                }
+                else if (!HasWritePermission(extraInformationCalculated))
```
MeniKadosh1: we need to remove also this write permission validation no? how it will validate path exist in Linux? try to add linux path and see if it allows you to proceed

PR #1291, opened 2019-09-17
Fixed the re-convert issue in legacy businessflow conversion

shahanemahesh:


Ginger/Ginger/Actions/ActionConversion/ActionsConversionWizard.cs
```diff
@@ -154,8 +154,8 @@ public async void ProcessConversion(ObservableList<BusinessFlowToConvert> lst, b
                     foreach (var bf in lst)
                     {
                         if (bf.IsSelected)
-                        {
-                            bf.BusinessFlow.RestoreFromBackup(true);
+                        {                            
+                            bf.BusinessFlow.RestoreFromBackup(clearBackup: false);
```
MeniKadosh1: please explain why we need to keep the backup?

Ginger/Ginger/Actions/ActionConversion/SelectBusinessFlowWzardPage.xaml.cs
```diff
@@ -74,6 +74,7 @@ private void SetActivitiesSelected()
         {
             foreach (var businessFlow in ListOfBusinessFlow)
             {
+                businessFlow.BusinessFlow.CreateBackup();
```
MeniKadosh1: 1- i don't think that under method called SetActivitiesSelected() we should do backup, it is not related, better call the method "PrepareBFsForConversion()"
2- why you iterate over all Activities and setting the SelectedForConversion = true, can't you as default convert all Activities if running in multi BF's conversion mode?

Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
```diff
@@ -495,10 +495,13 @@ private void RestoreList(string Name, IObservableList v, bool isLocalBackup = fa
             }
         }
 
-        public void RestoreFromBackup(bool isLocalBackup = false)
+        public void RestoreFromBackup(bool isLocalBackup = false, bool clearBackup = true)
         {
             RestoreBackup(isLocalBackup);
-            ClearBackup(isLocalBackup);
+            if (clearBackup)
```
MeniKadosh1: why this flag was added?

Ginger/Ginger/Actions/ActionConversion/ActionsConversionWizard.cs
shahanemahesh: we have kept the "Re-Convert" button visible and user can have the option to reconvert it multiple times, if suppose user have selected all the businessflows and reconvert them successfully and if user again clicks to reconvert button then there will not be any backup and nothing will be converted, if this is ok then we can remove the backup

Ginger/Ginger/Actions/ActionConversion/SelectBusinessFlowWzardPage.xaml.cs
shahanemahesh: 1 - if we want to create new method then will again have to iterate it again and will have to do backup, so just not to iterate it again added that line in that method.
2 - this is done since there is existing code in utils which is dependent on the "SelectedForConversion" flag, so will have to modify there as will this will require some tricky change so to avoid that added this method to set the flag for activities

Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
shahanemahesh: just added this flag if at some point user does not want to the clear backup then this flag can be used

Ginger/Ginger/Variables/VariabelsListViewPage.xaml.cs
```diff
@@ -140,11 +140,18 @@ private void ShowHideEditPage(VariableBase variabelToEdit)
                 }
                 else if (mVariabelsParent is BusinessFlow)
                 {
-                    mVariabelEditPage = new VariableEditPage(mVarBeenEdit, mContext, showAsReadOnly, VariableEditPage.eEditMode.BusinessFlow);
+                    mVariabelEditPage = new VariableEditPage(mVarBeenEdit, mContext, showAsReadOnly, VariableEditPage.eEditMode.Default);
                 }
                 else if (mVariabelsParent is Activity)
                 {
-                    mVariabelEditPage = new VariableEditPage(mVarBeenEdit, mContext, showAsReadOnly, VariableEditPage.eEditMode.Activity);
+                    if(mPageViewMode== General.eRIPageViewMode.View)
```
MeniKadosh1: what is that? why you modified VariableEditPage view mode?

shahanemahesh: i didnt do any change related to this, it might have came while merging from master (getting latest)

PR #1304, opened 2019-09-24
ProjectKey should be used while connecting with Jira, Domain was used…

shahanemahesh:
… so changed that

### DIFF
--- a/Ginger/GingerCore/ALM/JIRA/Bll/JiraImportManager.cs
+++ b/Ginger/GingerCore/ALM/JIRA/Bll/JiraImportManager.cs
@@ -60,7 +60,7 @@ namespace GingerCore.ALM.JIRA
                 if (resourceType == ResourceType.DEFECT)
                 {
                     AlmResponseWithData<JiraRepository.Data_Contracts.JiraFieldColl> testDefectFieldsList;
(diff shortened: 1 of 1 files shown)


Ginger/GingerCore/ALM/JIRA/Bll/JiraImportManager.cs
```diff
@@ -60,7 +60,7 @@ internal ObservableList<ExternalItemFieldBase> GetALMItemFields(ResourceType res
                 if (resourceType == ResourceType.DEFECT)
                 {
                     AlmResponseWithData<JiraRepository.Data_Contracts.JiraFieldColl> testDefectFieldsList;
-                    testDefectFieldsList = jiraRep.GetIssueFields(loginData.User, loginData.Password, loginData.Server, ALMCore.AlmConfig.ALMDomain, ALM_Common.DataContracts.ResourceType.DEFECT);
+                    testDefectFieldsList = jiraRep.GetIssueFields(loginData.User, loginData.Password, loginData.Server, ALMCore.AlmConfig.ALMProjectKey, ALM_Common.DataContracts.ResourceType.DEFECT);
```
MeniKadosh1: so if wrong field was used for connection how it was working when it was developed?

shahanemahesh: this is what i discussed with jinendra, but it might have happened that they might have the same domain name used as project key and so it was working, but as per me jira works on the project key to search any details from that particular project. And that is why it was just fetching the project details and not the more details

MeniKadosh1:
Merged it, please update related ticket

PR #1305, opened 2019-09-25
fix email report font & images

uzanh:


Ginger/Ginger/Reports/GingerExecutionReport/EmailExecutionReport.html
```diff
@@ -16,7 +16,7 @@
             padding: 0 !important;
             height: 100% !important;
             width: 100% !important;
-            font-family: CenturyGothic;
+            font-family: Arial, Helvetica, sans-serif;
```
MeniKadosh1: why we have this HTML duplicated on both Ginger and GingerCoreNET? 
i don't think we need the all "GingerExecutionReport" on Ginger project, we need it only on GingercoreNET- please confirm and if yes so delete it from Ginger

Ginger/GingerCoreNET/Reports/GingerExecutionReport/EmailExecutionReport.html
```diff
@@ -16,7 +16,7 @@
             padding: 0 !important;
             height: 100% !important;
             width: 100% !important;
-            font-family: CenturyGothic;
+            font-family: Arial, Helvetica, sans-serif;
```
MeniKadosh1: you need to do same changes also on "EmailExecutionReportOnLinux.HTML" which been used when running from Linux box

PR #1309, opened 2019-09-25
Changes related to Output values

aditijagtap:


Ginger/Ginger/Actions/ActionEditPage.xaml.cs
```diff
@@ -300,7 +300,6 @@ public void StopEdit()
         private void ReturnValues_CollectionChanged(object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e)
         {
             UpdateOutputTabVisual();
-            mAction.OnPropertyChanged(nameof(Act.ReturnValuesInfo));
```
JinendraGhodke: Need to test if during execution this icon is updated correctly

PR #1310, opened 2019-09-26
Handled the code to rollback on cancel button and disable the finish button

shahanemahesh:


Ginger/Ginger/Actions/ActionConversion/ActionsConversionWizard.cs
```diff
@@ -218,16 +223,33 @@ public async void BusinessFlowsActionsConversion(ObservableList<BusinessFlowToCo
...
+        /// This method will restore the conversion done
         /// </summary>
         public override void Cancel()
         {
-            base.Cancel();
+            try
+            {
+                if (IsConversionDoneOnce)
+                {
+                    foreach (BusinessFlowToConvert bfToConvert in ListOfBusinessFlow)
```
MeniKadosh1: try-Finally should be done as you did to insure the base.Cancel but try-catch for restore failure exception should be done around the single bfToConvert.BusinessFlow.RestoreFromBackup(); so if one will fail we still will try to restore others

shahanemahesh: Done

PR #1319, opened 2019-09-30
Unit Test added for "API Comparison Feature".

vyasprateek:


Ginger/GingerCoreTest/AutoPilot/APIModelsCompareTest.cs
```diff
@@ -0,0 +1,117 @@
...
+        [ClassInitialize]
+        public static void ClassInitialize(TestContext TC)
+        {
+            Reporter.ToLog(eLogLevel.DEBUG, "Creating the GingerCoreNET WorkSpace");
+            WorkSpaceEventHandler WSEH = new WorkSpaceEventHandler();
+            WorkSpace.Init(WSEH);
+
+            string TempSolutionFolder = TestResources.GetTestTempFolder(@"Solutions" + Path.DirectorySeparatorChar + "APIModelsComparisonUtilityTest");
+            if (Directory.Exists(TempSolutionFolder))
+            {
```
MeniKadosh1: 1- remove the dependency on external WSDL, instead save 5-6 APIModels on the Test Resources and load them into 1/2 lists in the Class Init to simulate ExsitingAPIModels LearnedAPIModels
2- instead 1 Unit test which test the all compare process, break it down into small unit tests which will test the building blucks functionalities like Match/Compare/etc. from each have several unit tests to test diffrent outcome

Ginger/GingerCoreTest/AutoPilot/APIModelsCompareTest.cs
```diff
@@ -0,0 +1,185 @@
...
+                Directory.Delete(TempSolutionFolder, true);
+            }
+
+            // Initialize XML Parser 
+            XMLTemplateParser xmlParser = new XMLTemplateParser();
+
+            // Initialize XML file names to be fetched from TestResources
+            xmlFiles = new List<string>()
+            {
+                @"XML\API_Templates_XML\Create_User.xml",
```
MeniKadosh1: why you saving the sample APIModels as regular xml's and not as APIModel xml's and then just load them with the NewRepositorySerializer?

Ginger/GingerCoreTest/AutoPilot/APIModelsCompareTest.cs
```diff
@@ -0,0 +1,185 @@
...
+                @"XML\API_Templates_XML\Delete_User.xml",
+                @"XML\API_Templates_XML\PhoneVerifySOAP_CheckPhoneNumber.xml",
```
MeniKadosh1: I don't see "API_Templates_XML" folder in the project files tree maybe that what causing the unit tests to fail because resource files failed to be found

Ginger/GingerCoreTest/AutoPilot/APIModelsCompareTest.cs
vyasprateek: Because this way we're testing the complete scenario of API Models Importing and Comparison mechanism. And I've updated the PR with fixes for failing on Azure.

Ginger/GingerCoreTest/AutoPilot/APIModelsCompareTest.cs
vyasprateek: Updated !

PR #1325, opened 2019-10-01
API Comparison Discussed changes and (Bug fix) 8037 msg box alert accept issue

vyasprateek:
Bug #8037 : Message Box Alert Accept Issue : FIXED

### DIFF
--- a/Ginger/GingerCore/Drivers/SeleniumDriver.cs
+++ b/Ginger/GingerCore/Drivers/SeleniumDriver.cs
@@ -840,71 +840,59 @@ namespace GingerCore.Drivers
 
         public override void RunAction(Act act)
         {
+            bool isActBrowser = act is ActBrowserElement;
+            ActBrowserElement actBrowserObj = isActBrowser ? (act as ActBrowserElement) : null;
+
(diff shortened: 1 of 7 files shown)


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -855,7 +855,12 @@ public override void RunAction(Act act)
                 if (act.GetType() == typeof(ActBrowserElement))
                 {
                     ActBrowserElement ABE = (ActBrowserElement)act;
-                    if (ABE.ControlAction == ActBrowserElement.eControlAction.SwitchToDefaultWindow)
+                    if (ABE.ControlAction == ActBrowserElement.eControlAction.AcceptMessageBox)
```
MeniKadosh1: try to just add IF that will check the action type + operation and if it one of those mentioned here so skip the rest of this method and jump to the DoRunAction.
if it works so please show to Jinendra to see if we missed any scenario.

MeniKadosh1:
Merged it, please update related tickets.

PR #1327, opened 2019-10-01
Bug fix/8032 pom i frame parent manual set is missing uneccessery switches

vyasprateek:
Now, user can add Parent IFrame for manually added elements in the POM. 
**Pending Unnecessary IFrame switching checks**


MeniKadosh1:
please update the relevant ticket

PR #1345, opened 2019-10-07
Issues Fixed 8133, 8134, 8137 and others, DataSource related 

vyasprateek:
8133:3.4-Getting error on clicking Compare API after selecting API
Bug 8134:3.4-Compare API-Unselected APIs are also compared
Bug 8137:Adding APIModel Action from SR is openeing the Action configuration Wizard


Ginger/Ginger/ApplicationModelsLib/APIModels/APIModelWizard/ScanAPIModelWizardPage.xaml.cs
```diff
@@ -88,7 +88,10 @@ void BtnCompareAPIClicked(object sender, RoutedEventArgs e)
             xApisSelectionGrid.btnMarkAll.Visibility = Visibility.Collapsed;
             xCompareBtnRow.Height = new GridLength(0);
 
-            xApisSelectionGrid.DataSourceList = AddAPIModelWizard.DeltaModelsList;
+            //Need to clear else if any item was selected on the Import Grid throws exception/error 
+            //as selected Item  won't anymore exist after updating the DataSource
+            AddAPIModelWizard.LearnedAPIModelsList.Clear();
+            xApisSelectionGrid.DataSourceList = new ObservableList<DeltaAPIModel>(AddAPIModelWizard.DeltaModelsList.OrderBy(d => d.comparisonStatus));
```
MeniKadosh1: but why you create new ObservableList<DeltaAPIModel> of Deta list before setting it as the grid DaraSource?

Ginger/Ginger/AutomatePageLib/AddActionMenu/ActionsFactory.cs
```diff
@@ -137,7 +137,7 @@ public static int AddActionsHandler(object mItem, Context mContext, int targetIn
         /// <returns></returns>
         static Act GenerateSelectedAction(Act selectedAction, Context mContext)
         {
-            if (selectedAction.AddActionWizardPage != null)
+            if (selectedAction.AddActionWizardPage != null && !selectedAction.ContainingFolder.Contains("SharedRepository"))
```
MeniKadosh1: why not to check check if "selectedAction.IsSharedRepositoryInstance == true"?? never go with validation on strings/names- only when must

Ginger/Ginger/ApplicationModelsLib/APIModels/APIModelWizard/ScanAPIModelWizardPage.xaml.cs
vyasprateek: We're _Ordering_ it based on the Comparison Status which returns _IOrderedEnumerable_ list of items while the DataSource is of type _IObservableList_

PR #1373, opened 2019-10-15
Bug fix/access data source related bugs on 11 oct2019 rc

vyasprateek:
8088 : 3.4- Compare API not working when adding the existing API
8097 : Runset : While running the viking flow from runset, execution getting stuck at perticular point.
8107 : 3.4- Delete All in LiteDB is not working for Key & Customized table
8154 : Access DataSource - While adding column to datasource it give exception
8158 : Access Datasource - Removing all rows gives exception
+ [Additional Issue] Always adding LiteDB even during an attempt of adding Access Data Source.


MeniKadosh1:
Merged, please update related bugs.

PR #1374, opened 2019-10-15
adding null check if matching property is null to avoid exception.

sudarshan0212:


Ginger/GingerCoreNET/Application Models/Delta/PomDelta/PomDeltaUtils.cs
```diff
@@ -294,7 +294,7 @@ private void SetMatchingElementDeltaDetails(ElementInfo existingElement, Element
                     latestProperty.Guid = matchingExistingProperty.Guid;
                     deltaProperty.ElementProperty = latestProperty;
                     if ((string.IsNullOrWhiteSpace(matchingExistingProperty.Value) == true && string.IsNullOrWhiteSpace(latestProperty.Value) == true)
```
JinendraGhodke: if the existing property value is null and on latest property value is not null.
Then deltastatus should be changed.

But with this condition it will be unchanged. which is not correct

PR #1391, opened 2019-10-18
Added the code to handle the Closing of Browser/Agent while recording is goingon

shahanemahesh:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -5115,6 +5116,11 @@ void AddJavaScriptToPage(string script)
                 string script3 = GetInjectJSSCript(script);
                 var v = ((IJavaScriptExecutor)Driver).ExecuteScript(script3, null);
             }
+            catch (OpenQA.Selenium.WebDriverException e)
+            {
+                Reporter.ToUser(eUserMsgKey.FailedToConnectAgent, mBrowserTpe, e.Message);
```
MeniKadosh1: 1- remove Reporter.ToUser()- we need to show toUser messgaed from UI pages and not from drivers/backend
2- improve exception Reporter.ToLog message to indicate what failed and not only the method name

shahanemahesh: Updated as per suggestion

PR #1392, opened 2019-10-18
Bug fix/8196 shared repo add to flow fix

aditijagtap:
Also added fix for Business flow delete all items operation.


Ginger/Ginger/AutomatePageLib/AddActionMenu/SharedRepositoryLib/ActionsRepositoryPage.xaml.cs
```diff
@@ -41,14 +41,14 @@ public partial class ActionsRepositoryPage : Page
     {
         readonly RepositoryFolder<Act> mActionsFolder;
         Context mContext;
-
-        public ActionsRepositoryPage(RepositoryFolder<Act> actionsFolder, Context context)
+        bool mIsAddToFlow=false;
```
MeniKadosh1: instead of adding another parameter "AddToFlow", notice that in case we load the sharedRepo page not from Automate page the "context" comes as Null, so if it comes as null do not add the AddToFlow button in tool bar- do it for all pages Actions/Activities/...

PR #1393, opened 2019-10-18
Added the code to handle "Ginger_Update" updation

shahanemahesh:


shahanemahesh:
This branch has the fix for 8193

Ginger/GingerCoreNET/ValueExpressionLib/ValueExpression.cs
```diff
@@ -656,7 +656,10 @@ public void ReplaceDataSource(string p)
                         string[] Name = tokens[1].Split(splitchar);
 
                         string[] markasdone = tokens[0].Split(new[] { "MASD=" }, StringSplitOptions.None)[1].Split(splitchar);
-
+                        if(!bDone)
```
MeniKadosh1: Need to test with 4 scenarioes:
2- scenarios when having DS as VE in some Action
2- scenarios when using Data Source Action

shahanemahesh: Hi Meni, checked the mentioned above scenario and it is working as expected

MeniKadosh1:
Merged, please update the ticket.

PR #1396, opened 2019-10-22
Fixed duplicate/copy runset

arvindgh:
added call to UpdateRunnersBusinessFlowRunsList function while saving runset as it making empty BusinessFlowRunsList while creating new Repo(like copy or duplicate)

### DIFF
--- a/Ginger/Ginger/RunSetPageLib/NewRunSetPage.xaml.cs
+++ b/Ginger/Ginger/RunSetPageLib/NewRunSetPage.xaml.cs
@@ -1043,6 +1043,8 @@ namespace Ginger.Run
                     runnerPage.RunnerPageEvent += RunnerPageEvent;
                     runnerPage.RunnerPageListener.UpdateBusinessflowActivities -= UpdateBusinessflowActivities;
                     runnerPage.RunnerPageListener.UpdateBusinessflowActivities += UpdateBusinessflowActivities;
+
+                    GR.IsUpdateBusinessFlowRunList = true;
                 });
             }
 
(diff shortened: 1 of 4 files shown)


Ginger/GingerCoreNET/Run/RunSetConfig.cs
```diff
@@ -177,7 +177,6 @@ public void UpdateRunnersBusinessFlowRunsList()
 
         public override void UpdateBeforeSave()
         {
-            UpdateRunnersBusinessFlowRunsList();
```
MeniKadosh1: why you removed the UpdateRunnersBusinessFlowRunsList() from UpdateBeforeSave()? with this change we will do the  UpdateRunnersBusinessFlowRunsList()  only in case runset is been saved from the specific save funtion in RunSetPage and not in any case as it should.
please explain why it was needed.

arvindgh: UpdateRunnersBusinessFlowRunsList() making RunnersBusinessFlowRunsList empty when we duplicating/ coping runset.
So UpdateRunnersBusinessFlowRunsList() is called from SaveRunSetConfig() to update the RunnersBusinessFlowRunList.

Ginger/Ginger/RunSetPageLib/NewRunSetPage.xaml.cs
```diff
@@ -1043,6 +1043,8 @@ private async Task<int> InitRunnersSection(bool runAsync = true, bool ViewMode=f
                     runnerPage.RunnerPageEvent += RunnerPageEvent;
                     runnerPage.RunnerPageListener.UpdateBusinessflowActivities -= UpdateBusinessflowActivities;
                     runnerPage.RunnerPageListener.UpdateBusinessflowActivities += UpdateBusinessflowActivities;
+
+                    GR.IsUpdateBusinessFlowRunList = true;
```
MeniKadosh1: this should be done inside RunsetExecutor.InitRunner() after BF's been loaded

MeniKadosh1:
Merged, please update the related ticket.

PR #1399, opened 2019-10-23
Fix for Runset Undo

aditijagtap:


Ginger/Ginger/RunSetPageLib/NewRunSetPage.xaml.cs
```diff
@@ -2498,9 +2498,9 @@ private void xUndoChangesBtn_Click(object sender, RoutedEventArgs e)
                 if (Ginger.General.UndoChangesInRepositoryItem(mRunSetConfig, true))
                 {
                     mRunSetConfig.SaveBackup();
+                    mRunSetConfig.GingerRunners.CollectionChanged += Runners_CollectionChanged;
+                    LoadRunSetConfig(mRunSetConfig, true);
                 }
```
MeniKadosh1: few items:
1- in which case the undo changes fails? why? what happens if it fails?
2- adding back the GingerRunners.CollectionChanged register must be done even if the backup failed because if not we will stop handling runners add/delete properly so need to check where to place if, maybe after the if unless it needed for LoadRunSetConfig() or maybe on Finally block with -= and then += to make sure we registering only once
3- if the undo fails we need to notify the user with Reporter.ToUser message

PR #1409, opened 2019-10-28
fix report with 2 actions issue

uzanh:


Ginger/GingerCoreNET/Run/RunListenerLib/TextFileRepository.cs
```diff
@@ -84,7 +84,7 @@ public override object SetReportAction(Act action, Context context, Amdocs.Ginge
                     try
                     {
                         screenShotCountPerAction++;
-                        if (executedFrom == Amdocs.Ginger.Common.eExecutedFrom.Automation)
+                        if (executedFrom == Amdocs.Ginger.Common.eExecutedFrom.Automation || Path.GetDirectoryName(action.ScreenShots[s]).EndsWith("LiteDBImages"))
```
MeniKadosh1: what is that? why we finding out if it is LiteDB via the path?

PR #1414, opened 2019-10-31
Feature/java pom implementation phase2

ravirk91:
Add Support for Update Elements
Move Switch Window action to   Legacy for All supported platforms and support conversion
Support Switch window in UI   element action
adding custom locator- Locator list is not filtered by Java locators


Ginger/GingerCore/Actions/ActSwitchWindow.cs
```diff
@@ -103,9 +103,19 @@ public Int32 WaitTime
             }
         }
 
+        private List<ePlatformType> obsoletePlatformList = new List<ePlatformType>()
```
JinendraGhodke: It is obsolete for all platforms for which it is supported. so why do we need to check?
We can return true no?

Ginger/GingerCore/Drivers/PBDriver/PBDriver.cs
```diff
@@ -96,7 +96,7 @@ public override void RunAction(Act act)
             mUIAutomationHelper.taskFinished = false;
             //TODO: avoid hard coded string... 
             actClass = actClass.Replace("GingerCore.Actions.", "");
-            if (actClass != "ActUIASwitchWindow" && actClass != "ActSwitchWindow")
+            if (actClass != "ActUIASwitchWindow" && actClass != "ActSwitchWindow" && (actClass == "Common.ActUIElement" && (((ActUIElement)act).ElementAction != ActUIElement.eElementAction.Switch)))
```
JinendraGhodke: use typeof instead of hardcoded class names which can easily break later

Ginger/GingerCore/Drivers/WindowsLib/WindowsDriver.cs
```diff
@@ -105,7 +105,16 @@ public override void RunAction(Act act)
 
             if (actClass != "ActSwitchWindow" || (actClass == "ActWindow" && ((ActWindow)act).WindowActionType != ActWindow.eWindowActionType.Switch))
             {
-                CheckRetrySwitchWindowIsNeeded();
+                var checkWindow = true;
+                if(actClass == "Common.ActUIElement" && (((ActUIElement)act).ElementAction != ActUIElement.eElementAction.Switch))
```
JinendraGhodke: use typof instead of name

Ginger/GingerCore/Drivers/WindowsLib/WindowsDriver.cs
```diff
@@ -105,7 +105,16 @@ public override void RunAction(Act act)
 
             if (actClass != "ActSwitchWindow" || (actClass == "ActWindow" && ((ActWindow)act).WindowActionType != ActWindow.eWindowActionType.Switch))
             {
-                CheckRetrySwitchWindowIsNeeded();
+                var checkWindow = true;
```
JinendraGhodke: why this flag? is it set to false for UI Element?

i think the retry is needed for a reason

PR #1425, opened 2019-11-05
Bug 8248 Fixed

vyasprateek:


Ginger/Ginger/BusinessFlowWindows/BusinessFlowConfigurationsPage.xaml.cs
```diff
@@ -158,8 +158,16 @@ public void UpdateBusinessFlow(BusinessFlow updateBusinessFlow)
             if (mBusinessFlow != updateBusinessFlow)
             {
                 ClearBindings();
+
```
MeniKadosh1: the remove of collectionChange and add back should be added to ClearBindings() and BindControls(); functions, this is why they were created

PR #1428, opened 2019-11-06
renamed the parameter to function

shahanemahesh:


Ginger/GingerCore/Actions/Rest/ActREST.cs
```diff
@@ -836,9 +836,9 @@ private string SetDynamicValues(ActREST AR, string ReqBody)
             return NewReqBody;
         }
 
-        bool IObsoleteAction.IsObsoleteForPlatform(ePlatformType Platfrom)
+        bool IObsoleteAction.IsObsoleteForPlatform(ePlatformType platform)
```
MeniKadosh1: please explain this fix:
1- on Interface the platform is with capital "P" so why you made it lower and why it is lower on all actions?
2- how this fix the show of new converted Action as described in bug 8220?

shahanemahesh: for 
#1 i will modify the parameter name in interface also
#2 There is spelling mistake "Platform" this variable is from "Act" class and the parameter to function is "Platfrom", currently in "ActRest" for this function "IsObsoleteForPlatform(ePlatformType Platfrom)" the condition used is "if(Platform == ePlatformType.WebServices || Platfrom == ePlatformType.NA)", so if you check the first condition "Platform == ePlatformType.WebServices" this is the variable "Platform" from the "Act" class and not the parameter to function is "Platfrom", so when this function is called first time the variable from "Act" class is set to some other platform and so for first time the action is not recognizing as Obsolete and so it is not converted but when businessflow is switched then this variable from the "Act" class is set correctly and so in "ActRest" class it works properly after businessflow is switched.

PR #1446, opened 2019-11-19
Fixed the issue in "MultiSetValue" "ActUIElement"

shahanemahesh:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -6556,7 +6556,7 @@ public void HandleActUIElement(ActUIElement act)
                         break;
 
                     case ActUIElement.eElementAction.MultiSetValue:
-                        List<IWebElement> textels = LocateElements(act.LocateBy, act.LocateValueCalculated);
+                        List<IWebElement> textels = LocateElements(act.ElementLocateBy, act.ElementLocateValue);
```
MeniKadosh1: if you will use "ElementLocateValue" and it will have VE inside (let's say variable) then here in driver you won't have the calculated value (Variable value).
you supposed to use "ElementLocateValueForDriver"- please change and test it with dynamic value.

MeniKadosh1:
Merged, please update ticket and user.

PR #1448, opened 2019-11-20
Updated the error message as per ticket

shahanemahesh:


Ginger/GingerCoreNET/ValueExpressionLib/ValueExpression.cs
```diff
@@ -560,7 +560,7 @@ public void ReplaceDataSource(string p)
                     }
                     if (dt.Rows.Count == 0)
                     {
-                        mValueCalculated = "No Row found with " + Query;
+                        mValueCalculated = "No row found with selected query";
```
MeniKadosh1: please explain what was the issue with showing the Query in the error message?

shahanemahesh: Query in error message doesnt look good so need to change the message as stated in bug

MeniKadosh1: i think we should keep the Query, only change error to "No row found with query: 'Select...'"

user should see the composed query so it will be easy for him to understand what is wrong in the query in case some parts of the query are dynamic

JinendraGhodke:
Please check if this PR is still relevant and update

PR #1463, opened 2019-11-27
Bug fix/8308 changes are not been saved when copying pasting tree items resources and b fs

vyasprateek:
Changes are lost for pasted/duplicated Tree Items on reloading solution : FIXED

The Dirty tracing for the recently added (pasted/duplicated) **tree item** doesn't gets initialized as we were updating their Dirty status to modified during the copy process.

Fix : setting dirty status to NoTrack in order to start their associated properties tracking and get the changes reflected once saved.


MeniKadosh1:
Merged, please update related ticket and user.

PR #1507, opened 2020-01-22
Fixed Bug 8379 Sending HTML Content using Free Text

arvindgh:
Removed IsBodyHTML = false for freeText to send html content through this


arvindgh:
> As discuss please test it for Send Free text action as well

Tested and Working as expected (as the flag  IsBodyHTML = true(default) send free email action is able to send plain text as well as html contents in the body, we are setting IsBodyHTML = false only for sms action).

PR #1508, opened 2020-01-22
Feature/java pom rocording support

ravirk91:


Ginger/GingerCore/Drivers/JavaDriverLib/JavaDriver.cs
```diff
@@ -2636,16 +2637,19 @@ public override void StartRecording()
...
         }
 
         void Amdocs.Ginger.Plugin.Core.IRecord.ResetRecordingEventHandler()
         {
             RecordingEvent = null;
         }
 
-        private void DoRecordings()
+        private bool isPOMRecording = false;
+        private void DoRecordings(bool learnAdditionalChanges = false)
```
JinendraGhodke, line 2649: why the flag name is learnAdditionalChanges? it should be isPOMRecording no ?

Ginger/GingerCore/Drivers/JavaDriverLib/JavaDriver.cs
```diff
@@ -2636,16 +2637,19 @@ public override void StartRecording()
...
         {
+            isPOMRecording = learnAdditionalChanges;
```
JinendraGhodke, line 2651: Need to set this flag to false in End Recording

PR #1512, opened 2020-01-28
Fix for UIElement IsVisible and IsEnabled function fix

aditijagtap:


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -6623,7 +6630,16 @@ public void HandleActUIElement(ActUIElement act)
                         MoveToElementActions(act);
                         break;
                     case ActUIElement.eElementAction.IsEnabled:
-                        act.AddOrUpdateReturnParamActual("Enabled", e.Enabled.ToString());
+                        if (e == null)
```
JinendraGhodke: Not required for isEnabled. E.g. if the element is not present then returning false is wrong

Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -6342,10 +6342,17 @@ public void HandleActUIElement(ActUIElement act)
 
             if (act.ElementLocateBy != eLocateBy.NA && (!act.ElementType.Equals(eElementType.Window) && !act.ElementAction.Equals(ActUIElement.eElementAction.Switch)))
             {
-                e = LocateElement(act);
-                if (e == null && act.ElementAction != ActUIElement.eElementAction.IsVisible)
+                if (act.ElementAction.Equals(ActUIElement.eElementAction.IsVisible) || (act.ElementAction.Equals(ActUIElement.eElementAction.IsEnabled)))
```
JinendraGhodke: Remove is Enabled check from here

PR #1519, opened 2020-02-07
Fixed FailedToConnectAgent while recording

arvindgh:
Added validation to check driver is running


Ginger/GingerCore/Drivers/SeleniumDriver.cs
```diff
@@ -5108,8 +5108,8 @@ void AddJavaScriptToPage(string script)
                 var v = ((IJavaScriptExecutor)Driver).ExecuteScript(script3, null);
             }
             catch (OpenQA.Selenium.WebDriverException e)
-            {                
-                StopRecordingIfAgentClosed();
+            {
+                StopRecordingIfAgentClosed(e.ToString());
```
JinendraGhodke: change it to e.Message()

PR #1587, opened 2020-04-06
Fixing TextLogger- Runset Execution & passrate cal. for Html Report

ravirk91:
Fixing TextLogger- Runset Execution & passrate cal. for Html Report


Ginger/GingerCoreNET/Reports/GingerExecutionReport.cs
```diff
@@ -2704,6 +2709,11 @@ public static string CreateGingerExecutionReport(ReportInfo RI, bool calledFromA
             switch (RI.reportInfoLevel)
             {
                 case ReportInfo.ReportInfoLevel.RunSetLevel:
+                    int totalRunners = WorkSpace.Instance.RunsetExecutor.Runners.Count;
```
JinendraGhodke, line 2712: Put all this inside 1 function and add try-catch. Because with current code if any exception then an entire report will come as blank.

PR #1614, opened 2020-05-21
Fixed calculate run description for activity variables

arvindgh:
value expression searches variables from current activity


arvindgh:
> Please test the run action scenario with lite DB and confirm

Working fine

PR #1619, opened 2020-05-28
Adding support to fetch widgets element attribute value

ravirk91:
Adding support to get widgets element attribute value


Ginger/GingerCore/Resources/GingerHTMLHelper.js
```diff
@@ -1271,6 +1275,13 @@ function define_GingerLib() {
         return pl;
     }
 
+    GingerLib.GetAttributeValue = function (el,attribute) {
+        var pl = new GingerPayLoad("Attribute Value");
+        pl.AddValueString(el.getAttribute(attribute).replace(/GingerHighlight/g,''));
```
JinendraGhodke, line 1280: if el.getAttribute will return null then .replace will throw exception. This will break the payload response. Please add null check to avoid exception

PR #1627, opened 2020-06-09
Fixed Ginger Testing Ginger-Variable UI- failed with many repeated errors

arvindgh:
If exception is thrown before the running action it should fail

### DIFF
--- a/Ginger/GingerCoreNET/Run/GingerRunner.cs
+++ b/Ginger/GingerCoreNET/Run/GingerRunner.cs
@@ -994,6 +994,11 @@ namespace Ginger.Run
             catch(Exception ex)
             {
                 Reporter.ToLog(eLogLevel.ERROR, "Exception in Run Action", ex);
+                if (string.IsNullOrEmpty(act.Error))
+                {
+                    act.Error = "Exception in Run Action " + ex;
+                    act.Status = eRunStatus.Failed;
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


Ginger/GingerCoreNET/Run/GingerRunner.cs
```diff
@@ -994,6 +994,11 @@ public void RunAction(Act act, bool checkIfActionAllowedToRun = true, bool moveT
             catch(Exception ex)
             {
                 Reporter.ToLog(eLogLevel.ERROR, "Exception in Run Action", ex);
+                if (string.IsNullOrEmpty(act.Error))
+                {
+                    act.Error = "Exception in Run Action " + ex;
```
JinendraGhodke, line 999: adding entire ex into Act.Error will add full stack strace to the error. And on Report it will break the layout. Please add only ex.message

Ginger/GingerCoreNET/Run/GingerRunner.cs
```diff
@@ -994,6 +994,11 @@ public void RunAction(Act act, bool checkIfActionAllowedToRun = true, bool moveT
...
+                    act.Status = eRunStatus.Failed;
```
JinendraGhodke, line 1000: We should not set Status explicitly. At the end of executiion there is function which caluclates actions status. If act.Error has some value it will mark it as failed

PR #1639, opened 2020-06-25
Fixed Unable to open failed Action

arvindgh:
Added exception handling for ProcessInputValueForDriver


Ginger/GingerCoreNET/Run/GingerRunner.cs
```diff
@@ -1423,44 +1423,50 @@ public void ProcessReturnValueForDriver(Act act)
...
+                                IV.ValueForDriver = string.Empty;
+                            }
                         }
                     }
                 }
+                act.ValueExpression.DecryptFlag = true;
+            }
+            catch (Exception ex)
+            {
+                Reporter.ToLog(eLogLevel.ERROR, "Exception occurred in ProcessInputValueForDriver : ", ex);
```
MeniKadosh1, line 1468: 1- so only try-catch was added? if yes, please explain why it's enough and not need to handle the exception reason

2- maybe need to add the try-catch for each calculation in the list so exception happened only the specific value will

MeniKadosh1, line 1468: The error message don't add anything, we know it's exception by adding ex object and we know it happen in ProcessInputValueForDriver from exception stackTrace which will be added to the log from ex object. Error message should be something like
Reporter.ToLog(eLogLevel.ERROR, string."Error occured while calculating the Action input VE's", ex);

Ginger/GingerCoreNET/ValueExpressionLib/ValueExpression.cs
```diff
@@ -96,7 +96,7 @@ public class ValueExpression : IValueExpression
         public static Regex rxEnvUrlPattern = new Regex(@"{(\bEnvURL App=)\w+\b[^{}]*}", RegexOptions.Compiled);
         
         private static Regex VBSRegex = new Regex(@"{[V|E|VBS]" + rxVar + "[^{}]*}", RegexOptions.Compiled);
-        private static Regex rxe = new Regex(@"{RegEx" + rxVare + ".*}", RegexOptions.Compiled);
+        private static Regex rxe = new Regex(@"{RegEx" + rxVare + ".*}", RegexOptions.Singleline);
```
MeniKadosh1, line 99: why "Singleline" option is insted "Compiled" option? can't we have both?
like "RegexOptions.Compiled | RegexOptions.Singleline"?

arvindgh: Yes, Updated.

Ginger/GingerCoreNET/ValueExpressionLib/ValueExpression.cs
```diff
@@ -96,7 +96,7 @@ public class ValueExpression : IValueExpression
         public static Regex rxEnvUrlPattern = new Regex(@"{(\bEnvURL App=)\w+\b[^{}]*}", RegexOptions.Compiled);
         
         private static Regex VBSRegex = new Regex(@"{[V|E|VBS]" + rxVar + "[^{}]*}", RegexOptions.Compiled);
-        private static Regex rxe = new Regex(@"{RegEx" + rxVare + ".*}", RegexOptions.Compiled);
+        private static Regex rxe = new Regex(@"{RegEx" + rxVare + ".*}", RegexOptions.Compiled | RegexOptions.Singleline);
         private static Regex rfunc = new Regex("{Function(\\s)*Fun(\\s)*=(\\s)*([a-zA-Z]|\\d)*\\((\")*([^\\)}\\({])*(\")*\\)}", RegexOptions.Compiled);
```
MeniKadosh1, line 100: please add unit tests, with and without new line to make sure Regex works as we expected

MeniKadosh1:
Arvind please check why Windows unit tests are failing

PR #1640, opened 2020-06-25
Fixed Hide DB Password in the log

arvindgh:


Ginger/GingerCore/Environments/Database.cs
```diff
@@ -494,7 +494,9 @@ public Boolean Connect(bool displayErrorPopup = false)
             }
             catch (Exception e)
             {
-                Reporter.ToLog(eLogLevel.ERROR, "DB connection failed, DB type: " + DBType.ToString() + "; Connection String =" + ConnectionString, e);
+                //Hide paasword from the ConnectionString
+                connectConnectionString = connectConnectionString.Replace(connectConnectionString.Substring(connectConnectionString.ToLower().IndexOf("password") + "password".Length), "=*****");
```
MeniKadosh1, line 498: 1-This code change assum "password" will be last in the connection string, but is it always? can't password come before userName or other?

2- check how password appera for all common connection strings, for example for MySQL I see the password is "Pwd"
sample: "Server=myServerAddress;Database=myDataBase;Uid=myUsername;Pwd=myPassword;"

one approach can be to remove all spaces from connection string and then look for for "password=" or "pwd=" (or other password string types if there is) end index and then to replace all between this index to ";" or end of string with "****"

arvindgh: Done. Please review

Ginger/GingerCore/GeneralLib/General.cs
```diff
@@ -906,6 +906,43 @@ public static T TryFindParent<T>(DependencyObject child)
...
+        /// This function is used to replace the password value from the string with *****
+        /// </summary>
+        /// <param name="dataString">The string argument which has password value</param>
+        /// <returns></returns>
+        public static string HidePasswordFromString(string dataString)
+        {
+            string passwordValue = dataString.Replace(" ", "");//remove spaces
+            string passwordString = string.Empty;
+            //Matching string
+            if (dataString.ToLower().Contains("pwd="))
```
MeniKadosh1, line 920: did you confirmed that no exception will be trowen in case the string do not contains both "Pwd" or "Password"?

arvindgh: Returning original string if it does not contain matching string("pwd", or "password")

Ginger/GingerCore/GeneralLib/General.cs
```diff
@@ -906,6 +906,48 @@ public static T TryFindParent<T>(DependencyObject child)
...
+            {
+                passwordString = "pwd=";
+            }
+            else if (dataString.ToLower().Contains("password="))
+            {
+                passwordString = "password=";
+            }
+            else
+            {
```
MeniKadosh1, line 929: please add unit tests for all types on connection strings password/pwd combinations

Ginger/GingerCore/GeneralLib/General.cs
```diff
@@ -906,6 +906,48 @@ public static T TryFindParent<T>(DependencyObject child)
             else
                 return TryFindParent<T>(parentObject);
         }
+
+        /// <summary>
+        /// This function is used to replace the password value from the string with *****
+        /// </summary>
```
MeniKadosh1, line 912: need to mention it also will remove all spaces... that probably the issues with this method to be generic for now

arvindgh: We are not removing password from the original connection string, but we will face issue when we have space in the password itself

Ginger/GingerCore/GeneralLib/General.cs
arvindgh: Done

Ginger/GingerCoreTest/Misc/DataBaseTest.cs
```diff
@@ -362,5 +362,53 @@ public void RenameDBName()
             Assert.AreEqual(actDB.DBName, db.Name, "Names");
 
         }
+
+        [TestMethod]
+        [Timeout(60000)]
+        public void TestHidePasswordFromString()
+        {
+            //Access Db
+            string ConString = "Provider=Microsoft.ACE.OLEDB.12.0;Data Source=C:/myAccessFile.accdb;Jet OLEDB:Database Password=MyDbPassword;";
```
MeniKadosh1, line 371: correct way of writing unit test should include the 3 A's, //Arrange, //Act and //Assert:
//Arrange
here you should place the preperation code, like:
string ConString = "Provider=Microsoft.ACE.OLEDB.12.0;Data Source=C:/myAccessFile.accdb;Jet OLEDB:Database Password=MyDbPassword;";

//Act
here you place the code you testing, like:
string Result = General.HidePasswordFromString(ConString);

//Assert
here, you add all asserts, like:
Assert.AreNotEqual(ConString, Result);
Assert.AreEqual(Result.Contains("Password=*****"), true);

please re desing the test with the 3 A's, you also can/should split it to unit test per connection string type

Ginger/GingerCoreTest/Misc/DataBaseTest.cs
```diff
@@ -362,5 +362,53 @@ public void RenameDBName()
...
+            string Result = General.HidePasswordFromString(ConString);
+            Assert.AreNotEqual(ConString, Result);
+            Assert.AreEqual(Result.Contains("Password=*****"), true);
```
MeniKadosh1, line 374: for all Asserts, please add also assert which make sure the actual password was removed, like:
Assert.AreEqual(Result.Contains("MyDbPassword"), false);

PR #1643, opened 2020-07-02
Removed default proxy values

arvindgh:


Ginger/GingerCore/Drivers/Mobile/Perfecto/PerfectoDriver.cs
```diff
@@ -78,7 +78,7 @@ public class PerfectoDriver : DriverBase
 
         [UserConfigured]
         [UserConfiguredDefault("")]
-        [UserConfiguredDescription("Proxy Settings, if set, Proxy will be considered otherwise it will go to system proxy e.g 'genproxy.amdocs.com:8080'")]
+        [UserConfiguredDescription("Proxy Settings, if set, Proxy will be considered otherwise it will go to system proxy e.g '[sub-domain.]example.com:8080'")]
```
JinendraGhodke, line 81: what is '[sub-domain.]example.com ? is it some standard proxy format? To me, it looks confusing

arvindgh: Yes. we provide optional values in [ ].

JinendraGhodke: In that case, it should be [sub-domain].example.com 
You put the dot inside bracket which is causing confusion

arvindgh: Dot is also optional, it will only come when we have subdomain e.g.
- example.com
- abc.example.com

PR #1647, opened 2020-07-07
Fixed Check-In Solution Changes

arvindgh:
RepositoryFolderBase object can be null in case of solution file


Ginger/GingerCoreNET/SourceControl/SourceControlIntegration.cs
```diff
@@ -70,14 +70,14 @@ public static bool UpdateFile(SourceControlBase SourceControl, string path)
             string error = string.Empty;
             bool IsFileUpdated = true;
             RepositoryFolderBase repositoryFolderBase = WorkSpace.Instance.SolutionRepository.GetRepositoryFolderByPath(Path.GetDirectoryName(path));
-            repositoryFolderBase.PauseFileWatcher();
+            repositoryFolderBase?.PauseFileWatcher();
```
JinendraGhodke, line 73: In which cases the foldebase object will be null ?

arvindgh:
> In which cases the foldebase object will be null ?

repositoryFolderBase can be null in case of Solution.xml file

PR #1666, opened 2020-08-07
Performance improvement/post merged changes test and update

vyasprateek:
Default BusinessFlow selected on Runner loads Issue Fixed
Undo RepoItem Issue Fixed
Child Items with New Guid Issue Fixed
more Unit Tests added covering new functionalities

### DIFF
--- a/Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
+++ b/Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
@@ -579,17 +579,24 @@ namespace Amdocs.Ginger.Repository
 
         public bool RestoreFromBackup(bool isLocalBackup = false, bool clearBackup = true)
         {
+            bool isRestored = false;
             try
             {
-                return RestoreBackup(isLocalBackup);
+                isRestored = RestoreBackup(isLocalBackup);
+            }
+            catch(Exception exc)
+            {
+                Reporter.ToLog(eLogLevel.DEBUG, "Error occured in the Undo Process", exc);
             }
             finally
             {
-                if (clearBackup)
+                if (isRestored && clearBackup)
                 {
                     ClearBackup(isLocalBackup);
                 }
             }
+
+            return isRestored;
         }
 
         private string mFileName = null;
--- a/Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
+++ b/Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
@@ -670,7 +670,7 @@ namespace Amdocs.Ginger.Repository
             return dt2;
         }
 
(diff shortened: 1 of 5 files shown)


Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
```diff
@@ -579,17 +579,24 @@ private void RestoreList(string Name, IObservableList v, bool isLocalBackup = fa
...
         {
+            bool isRestored = false;
             try
             {
-                return RestoreBackup(isLocalBackup);
+                isRestored = RestoreBackup(isLocalBackup);
+            }
+            catch(Exception exc)
+            {
+                Reporter.ToLog(eLogLevel.DEBUG, "Error occured in the Undo Process", exc);
```
MeniKadosh1, line 589: this should not be "Debug" error but "Error" level, "Debug" is only for extra info for the developer- here we have Error which needs to be recorded

Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
```diff
@@ -714,19 +721,25 @@ private RepositoryItemBase CopyRIObject(RepositoryItemBase repoItemToCopy)
                     Reporter.ToLog(eLogLevel.ERROR, string.Format("Error occured during object copy of the item: '{0}', type: '{1}', property/field: '{2}'", this.ItemName, this.GetType(), mi.Name), ex);
                 }
             });
+            
+            if(SetNewGUID)
```
MeniKadosh1, line 725: that is not enough, there is a reason for the GUID replace which been done using ser/des- this is because we need to replace also GUID's placed in different places as String (like in Flow Control- GoToAction)- please undo all changes and keep the GUID replace as before.

Ginger/GingerCoreNETUnitTest/SolutionRepositoryLib/RepositoryTest.cs
```diff
@@ -1281,5 +1281,136 @@ public void CopyRepoItem_TrackCopyTest()
...
+            Activity activity = new Activity();
+            activity.ActivityName = "Login";
+
+            ActDummy dumAct = new ActDummy();
+            //Guid sfGuid = Guid.NewGuid();
+            FlowControl sampleFC = new FlowControl()
+            {
+                Active = true,
+                FlowControlAction = eFlowControlAction.GoToAction,
+                ConditionCalculated = "aaa",
```
MeniKadosh1, line 1300: this is not enough- need to set the Selected Action value- which is been done with GUID#Name combination- see how it saved to XML when you doing it from UI and mimic it on unit test- then need to validate the GUID of poined Action to see if it was changed/ unchanged as expected based on Copy with new GUID or copy with same GUID

PR #1681, opened 2020-08-24
Bug fix/7318 add output value in api model issue

ravirk91:
Bug fix/7318 add output value in api model issue


JinendraGhodke (commented):
Please check why build is failing

JinendraGhodke (approved):

PR #1692, opened 2020-09-14
virtual driver fixes

ambarkhan:


ambarkhan:
failures due to rest service down for thr test

PR #1789, opened 2021-01-19
Connection page - zephyr integration to ginger

uzanh:


Ginger/GingerCore/ALM/ZephyrEntCore.cs
```diff
@@ -0,0 +1,338 @@
...
+        }
+
+        public TestCaseResource CreateTestCase(long treeNodeId, ActivitiesGroup activtiesGroup)
+        {
+            return zephyrEntExportManager.CreateTestCase(treeNodeId, activtiesGroup);
+        }
+
+        public override Dictionary<Guid, string> CreateNewALMDefects(Dictionary<Guid, Dictionary<string, string>> defectsForOpening, List<ExternalItemFieldBase> defectsFields, bool useREST = false)
+        {
+            throw new NotImplementedException();
```
JinendraGhodke, line 109: Does Zephyr support defects creation ? if yes we need to implement this feature like we have for other ALMs

Ginger/GingerCoreNET/ALMLib/ALMUserConfig.cs
```diff
@@ -64,6 +64,20 @@ public string ALMPassword
                 EncryptedALMPassword = EncryptionHandler.EncryptString(value, ref res);
             }
         }
+        private string mALMToken;
+        [IsSerializedForLocalRepository]
```
JinendraGhodke, line 68: Can we use the password field itself for token instead of new field ?

Ginger/GingerCoreCommon/ALM/ALMIntegration.cs
```diff
@@ -29,7 +29,8 @@ public enum eALMType
             RALLY = 3,
             Jira = 4,
             Qtest = 5,
-            Octane = 6
+            Octane = 6,
+            ZephyrEnt = 7
```
JinendraGhodke, line 33: Please use full name ZephyrEnterprise instead of Abbreviation

Ginger/GingerCore/ALM/ZephyrEntCore.cs
```diff
@@ -0,0 +1,338 @@
...
+    public class ZephyrEntCore : ALMCore
+    {
+        protected Zepyhr_Ent_Repository.ZephyrEntRepository zephyrEntRepository;
+        private ZephyrEntExportManager zephyrEntExportManager;
+        private ZephyrEntImportManager zephyrEntImportManager;
+        public override ObservableList<ActivitiesGroup> GingerActivitiesGroupsRepo { get; set; }
+        public override ObservableList<Activity> GingerActivitiesRepo { get; set; }
+        public ZephyrEntCore()
+        {
+            string settingsPath = DefaultAlmConfig.ALMConfigPackageFolderPath;
```
JinendraGhodke, line 33: settingspath is unused here. Please remove if not required

JinendraGhodke (changes requested):

Ginger/GingerCoreNET/ALMLib/ALMUserConfig.cs
uzanh: i asked to add the token option and to keep the password option

Ginger/GingerCore/ALM/ZephyrEntCore.cs
uzanh: done

Ginger/GingerCore/ALM/ZephyrEntCore.cs
uzanh: no, created with jira

Ginger/GingerCoreCommon/ALM/ALMIntegration.cs
uzanh: done

Ginger/GingerCoreNET/ALMLib/ALMUserConfig.cs
JinendraGhodke: Yes. For sure we need to keep the password option as well. But what i am saying is to use the same field i,e, ALMPassword to store the token on XML. Because user can use only 1 option at a time

Also when running from command line we are using the password field. if we want to add a new field then need to add it also in CLI

Ginger/Ginger/ALM/ALMConnectionPage.xaml.cs
```diff
@@ -653,14 +634,18 @@ private void ALMRadioButton_Checked_Changed(object sender, RoutedEventArgs e)
...
+
                     default:
                         //Not used
                         break;
                 }
                 //Clear bindings
                 BindingOperations.ClearAllBindings(ServerURLTextBox);
                 BindingOperations.ClearAllBindings(RestAPICheckBox);
-                BindingOperations.ClearAllBindings(UserNameTextBox);
+                BindingOperations.ClearAllBindings(ZephyrEntTokenCheckBox);
```
MeniKadosh1, line 648: need to bring back 'BindingOperations.ClearAllBindings(UserNameTextBox);'

MeniKadosh1 (approved):

PR #1791, opened 2021-01-20
Octance issue fixes

Maheshkale447:
Fix for below 2 issues
1. Run suite status fix.
2. Defect field was not getting updated properly.


Ginger/GingerCore/ALM/OctaneCore.cs
```diff
@@ -1,20 +1,4 @@
-#region License
-/*
-Copyright © 2014-2020 European Support Limited
```
JinendraGhodke, line 1: Copy right license is required. Please add it back

Maheshkale447: Added license again

JinendraGhodke (approved):

PR #1830, opened 2021-03-18
Bug fix/8933 failed to export execution details qc

arvindgh:


JinendraGhodke (commented):
if it is for beta blocker then branch needs to be created from beta and merge into beta

arvindgh:
> if it is for beta blocker then branch needs to be created from beta and merge into beta

yes created from beta, by mystically mapped to master updated now

JinendraGhodke (approved):

PR #1861, opened 2021-04-11
Features/action file transfer2 linux transformation

uzanh:


Ginger/Ginger/Actions/ActionEditPages/ActFileTransferEditPage.xaml.cs
```diff
@@ -71,15 +71,15 @@ private void BrowsePCPathButton_Click(object sender, RoutedEventArgs e)
...
             if (dlg.ShowDialog() == System.Windows.Forms.DialogResult.OK)
             {
                 // replace Absolute file name with relative to solution
-                string FileName = dlg.FileName.ToUpper();
-                if (FileName.Contains(SolutionFolder))
+                string FileName = dlg.FileName;
+                if (FileName.ToUpper().Contains(SolutionFolder.ToUpper()))
                 {
-                    FileName = FileName.Replace(SolutionFolder, @"~\");
+                    FileName = @"~\" + FileName.Remove(0, SolutionFolder.Length);
```
MeniKadosh1, line 82: why you preferred Remove than replace?

Let's use centalized function (need to check if exist and if not so to add one) which deals with relative path handling instead doing it inside the Action code

Ginger/GingerCore/GingerCore.csproj
```diff
@@ -885,7 +884,7 @@
       <Version>1.8019.4006.503</Version>
     </PackageReference>
     <PackageReference Include="SSH.NET">
-      <Version>2016.1.0</Version>
+      <Version>2020.0.1</Version>
```
MeniKadosh1, line 887: did you tested all possible impact due to upgrade also on Unix driver and other depended code?

Ginger/Ginger/Actions/ActionEditPages/ActFileTransferEditPage.xaml.cs
uzanh: 1. not to be depend on case sensitive
2. ok, we have one

Ginger/GingerCore/GingerCore.csproj
uzanh: check agent , what we have more?

MeniKadosh1 (approved):

PR #1876, opened 2021-04-21
Windows UIElementAction operations

arvindgh:


Ginger/GingerCore/Platforms/PlatformsInfo/WindowsPlatform.cs
```diff
@@ -167,6 +204,7 @@
                 mElementsTypeList.Add(eElementType.List);
                 mElementsTypeList.Add(eElementType.Table);
```
JinendraGhodke, line 205: Remove the table element type as it is not supported for windows

JinendraGhodke (changes requested):

JinendraGhodke (approved):

PR #1879, opened 2021-04-22
Feature/self healing 1

ravirk91:
Self-healing adding support to prioritize locator position during POM execution


JinendraGhodke (commented):
As discussed please add the code for selenium driver as well we will merge it

JinendraGhodke (approved):

PR #1880, opened 2021-04-26
Added support to map target platform while adding Activity or Activities group from shared repository

Maheshkale447:
Added support to map target platform while adding Activity or Activities group from shared repository to current BF


Ginger/Ginger/AutomatePageLib/AddActionMenu/ActionsFactory.cs
```diff
@@ -269,10 +270,29 @@ public static void AddActivitiesFromSRHandler(List<Activity> sharedActivitiesToA
 
             if (parentGroup != null)
             {
+
                 foreach (Activity sharedActivity in sharedActivitiesToAdd)
                 {
+                    eUserMsgSelection userSelection = eUserMsgSelection.None;
                     Activity activityIns = (Activity)sharedActivity.CreateInstance(true);
                     activityIns.Active = true;
+                    //map activities target application to BF if missing in BF
```
JinendraGhodke, line 279: This code to map activities TA to BF is duplicated in 2 places. Can we move it to common method and call this method ?

Ginger/Ginger/AutomatePageLib/AddActionMenu/ActionsFactory.cs
```diff
@@ -289,10 +309,30 @@ public static void AddActivitiesGroupsFromSRHandler(List<ActivitiesGroup> shared
         {
             foreach (ActivitiesGroup sharedGroup in sharedActivitiesGroupsToAdd)
             {
+                eUserMsgSelection userSelection = eUserMsgSelection.None;
                 ActivitiesGroup droppedGroupIns = (ActivitiesGroup)sharedGroup.CreateInstance(true);
                 businessFlow.AddActivitiesGroup(droppedGroupIns);
                 ObservableList<Activity> activities = WorkSpace.Instance.SolutionRepository.GetAllRepositoryItems<Activity>();
-                businessFlow.ImportActivitiesGroupActivitiesFromRepository(droppedGroupIns, activities, false);
+                foreach (Activity activityIns in activities)
+                {
```
JinendraGhodke, line 317: Inside ImportActivitiesGroupActivitiesFromRepository we must be iterating all activities. So better to check TA there and call the method that you will create from there instead of iterating twice

JinendraGhodke (changes requested):

Ginger/Ginger/AutomatePageLib/AddActionMenu/ActionsFactory.cs
Maheshkale447: Created function to map TA to BF for duplicate code

JinendraGhodke (approved):

PR #1898, opened 2021-05-10
Bug fix/beta bug 8992

ravirk91:
fixing recording issue with POM


JinendraGhodke:
Need to be handled differently

PR #1909, opened 2021-05-12
Feature/enhancing explorer functionalities

vyasprateek:
--> LiveSpy page now displays new UCElementDetails 

Below Bugs Fixed;-

[Bug 8993](http://ilrnaginger02:8080/tfs/Automation/Ginger/_workitems/edit/8993):
Java POM: Locator is not active while adding Element using Spy

[Bug 9003](http://ilrnaginger02:8080/tfs/Automation/Ginger/_workitems/edit/9003):
Window Explorer-Mobile- Duplicate ByXpath Locators with wrong data

[Bug 9009](http://ilrnaginger02:8080/tfs/Automation/Ginger/_workitems/edit/9009):
Window Explorer-Mobile- No need Window/Page selection control for Mobile

[Bug 9026](http://ilrnaginger02:8080/tfs/Automation/Ginger/_workitems/edit/9026):
Window Explorer- Add action adding multiple actions

[Bug 9027](http://ilrnaginger02:8080/tfs/Automation/Ginger/_workitems/edit/9027):
Windows Explorer- Possible Memory leak on this page

[Bug 9028](http://ilrnaginger02:8080/tfs/Automation/Ginger/_workitems/edit/9028):
Window Explorer -Run button running the action as well as giving error

### DIFF
--- a/Ginger/GingerCore/Drivers/ASCF/ASCFDriver.cs
+++ b/Ginger/GingerCore/Drivers/ASCF/ASCFDriver.cs
@@ -1481,6 +1481,11 @@ namespace GingerCore.Drivers.ASCF
             return false;
         }
 
+        public bool IsWinowSelectionRequired()
+        {
+            return false;
+        }
+
         public List<eTabView> SupportedViews()
         {
             return new List<eTabView>() { /*eTabView.Screenshot, eTabView.GridView, eTabView.PageSource, eTabView.TreeView */};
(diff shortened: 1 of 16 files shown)


Ginger/GingerCore/Drivers/ASCF/ASCFDriver.cs
```diff
@@ -1481,6 +1481,11 @@ public bool IsLiveSpySupported()
             return false;
         }
 
+        public bool IsWinowSelectionRequired()
```
MeniKadosh1, line 1484: I don't think this supposed to be part of the Driver interface but part of the PlatformInfo.
Driver interface should include operations which Driver needs to do, like GetElementInfo(), but things like, titles to use, or if to show this configuration or another needs to be on PlatformInfo unless very specific to Driver type inside same Platform. 
Need to change in next PR

MeniKadosh1 (approved):

PR #1972, opened 2021-06-25
Feature/self healing custom relative xpath support

ravirk91:
Feature/self healing custom relative xpath support


Ginger/GingerCoreNET/Application Models/Common/CustomRelativeXpathTemplate.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return mStatus;
+            }
+            set
+            {
+                mStatus = value;
+                OnPropertyChanged(nameof(Value));
+                OnPropertyChanged(nameof(Status));
+            }
+        }
```
JinendraGhodke, line 62: why we need 2 separate fields if both are pointing to same value ? we can have only 1 field and use it

JinendraGhodke (changes requested):
Check my comment

JinendraGhodke (approved):

PR #1978, opened 2021-07-07
handling exception when injecting Ginger live spy

ravirk91:
handling exception when injecting Ginger live spy


JinendraGhodke (changes requested):
Add user message if failed to start live spy listener

JinendraGhodke (approved):

PR #1981, opened 2021-07-08
Fixing switch frame during POM widget execution

ravirk91:
Fixing switch frame during POM widget execution


Ginger/GingerCore/Drivers/JavaDriverLib/JavaDriver.cs
```diff
@@ -543,7 +543,7 @@ private void IntializeIfWidgetsElement(ElementInfo currentPOMElementInfo)
 
                     if (!string.IsNullOrEmpty(currentPOMElementInfo.Path))
                     {
-                        PayLoad PLSwitchFrame = new PayLoad("HTMLElementAction", "SwitchFrame", eLocateBy.ByXPath.ToString(), currentPOMElementInfo.XPath, string.Empty);
+                        PayLoad PLSwitchFrame = new PayLoad("HTMLElementAction", "SwitchFrame", eLocateBy.ByXPath.ToString(), currentPOMElementInfo.Path, string.Empty);
```
JinendraGhodke, line 546: Retrieve the parent Iframe path from properties and use that

JinendraGhodke (changes requested):
Retrieve the parent Iframe path from properties and use that

JinendraGhodke (approved):

PR #1988, opened 2021-07-19
Feature/self healing cofiguration

ravirk91:
Feature/self healing configuration


Ginger/GingerCoreNET/Drivers/CoreDrivers/Web/Selenium/SeleniumDriver.cs
```diff
@@ -3174,6 +3174,12 @@ public IWebElement LocateElement(Act act, bool AlwaysReturn = false, string Vali
                             SwitchFrame(currentPOMElementInfo);
                         }
                         elem = LocateElementByLocators(currentPOMElementInfo.Locators);
+
+                        if (pomExcutionUtil.AutoUpdateCurrentPOM() != null)
```
JinendraGhodke, line 3178: if element is not null then no need to update the pom. So we can add check elem== null then only go inside

Ginger/GingerCoreCommon/SelfHealingLib/SelfHealingConfig.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        [IsSerializedForLocalRepository]
+        public bool AutoFixAnalyzerIssue { get; set; }
+
+        [IsSerializedForLocalRepository]
+        public bool AutoUpdateApplicationModel { get; set; }
+
+        public bool SaveChangesInSourceControl { get; set; }
+
+        public bool RunFromAutomateTab { get; set; }
```
JinendraGhodke, line 38: we should not have this flag here. There is already existing flag for this. need to find and use the same flag for checking this

Ginger/GingerCoreNET/Application Models/Execution/POM/POMExecutionUtils.cs
```diff
@@ -122,5 +139,48 @@ public bool PriotizeLocatorPosition()
...
+        {
+            var passedLocator = GetCurrentPOMElementInfo().Locators.Where(x => x.LocateStatus == ElementLocator.eLocateStatus.Passed);
+
+            var pomLastUpdatedTimeSpan = (System.DateTime.Now -  GetCurrentPOM().RepositoryItemHeader.LastUpdate).TotalHours;
+            if (passedLocator == null || passedLocator.Count() > 0 || pomLastUpdatedTimeSpan < 5 )
+            {
+                return null;
+            }
+            
+            var deltaElementInfos = GetUpdatedVirtulPOM();
```
JinendraGhodke, line 153: what if self healing is disabled ?

JinendraGhodke (changes requested):

Ginger/GingerCoreCommon/UIElement/ElementInfo.cs
```diff
@@ -598,5 +643,13 @@
...
     }
 
-    
+    public enum SelfHealingInfoEnum
+    {
+        [EnumValueDescription(" ")]
+        None,
+        [EnumValueDescription("Element not found during self healing operation")]
+        ElementDeleted,
+        [EnumValueDescription("Element updated during self healing operation")]
```
JinendraGhodke, line 652: Element details auto-updated using self-healing

JinendraGhodke (changes requested):

JinendraGhodke (approved):

PR #2008, opened 2021-08-03
Bug fix/9127 duplicate business flow keep sr link for child items

JinendraGhodke:


Ginger/GingerCore/Repository/RepositoryItem.cs
```diff
@@ -118,57 +118,7 @@ public void InvokPropertyChanngedForAllFields()
...
-            // Make sure we must impl or get exception
-            throw new Exception("Unknown Type for Short Type Name " + T.Name);
-        }
-
-        public new string ObjFileExt { get { return FileExt(this.GetType()); } }
-
-        public new static string FolderName(Type T)
-        {
-            string s = GetShortType(T) + "s";
-            if (s.EndsWith("ys"))
```
MeniKadosh1, line 160: how you handled the "ies" if needed, or the handling of @"SharedRepository\" below?

Ginger/GingerCore/Repository/RepositorySerializer.cs
```diff
@@ -935,49 +935,17 @@ public object DeserializeFromFileObj(Type type, string fileName)
...
-            //TODO: more safe to use type of then Full Name - fix it!
-            if (t == typeof(BusinessFlow)) { return "BusinessFlow"; }
-            if (t == typeof(ActivitiesGroup)) { return "ActivitiesGroup"; }
-            if (t == typeof(Activity)) { return "Activity"; }
-            if (t == typeof(ErrorHandler)) { return "Activity"; }
-            if (t == typeof(CleanUpActivity)) { return "Activity"; }
-            if (typeof(Act).IsAssignableFrom(t)) { return "Action"; }
-            if (typeof(VariableBase).IsAssignableFrom(t)) { return "Variable"; }
-            if (typeof(DataSourceBase).IsAssignableFrom(t)) return "DataSource";
-            if (t.FullName == "GingerCore.Agent") return "Agent";
```
MeniKadosh1, line 964: any idea why some goes by type and some by FullName?

Ginger/GingerCore/Repository/RepositoryItem.cs
JinendraGhodke: FolderName Method is only used in get of ObjFolderName. And ObjFolderName have 0 references. 

Also we have same field on RepositoryItemBase. It is handled from there.

Ginger/GingerCore/Repository/RepositorySerializer.cs
JinendraGhodke: Not sure. will cross-check if it makes any difference.

JinendraGhodke: Verified that there is no impact of the different ways we are checking type

PR #2016, opened 2021-08-11
Mobile fixes/fixes and enhancements

vyasprateek:


Ginger/GingerCoreNET/Drivers/CoreDrivers/Mobile/Appium/GenericAppiumDriver.cs
```diff
@@ -415,6 +415,12 @@ public override void RunAction(Act act)
                     return;
                 }
 
+                if(actionType == typeof(ActGotoURL))
+                {
+                    Reporter.ToUser(eUserMsgKey.StaticInfoMessage, "Navigating to Native Application not supported");
```
MeniKadosh1, line 420: i didn't got this solution:
1- showing user message from driver is wrong, showing user msg should be initiate only from UI pages
2- no need to check by Act type, need to decide this by platform info or other
3- i didn't saw this message is been shown only if Agent is Mobile Native, it will be shown for Mobile Web as well no? while for web we can do it

MeniKadosh1 (approved):

PR #2032, opened 2021-08-17
implementing code to validate if execution is duplicated

sudarshan0212:


Ginger/GingerCoreNET/Run/RunListenerLib/CenteralizedExecutionLogger/AccountReportApiHandler.cs
```diff
@@ -222,6 +223,30 @@ public async Task SendActionExecutionDataToCentralDBAsync(AccountReportAction ac
...
+        public bool ExecutionIdValidation(Guid executionId)
+        {
+            RestRequest restRequest = (RestRequest)new RestRequest(EXECUTION_ID_VALIDATION + executionId, Method.GET);           
+            string message = string.Format("execution id : {0}", executionId);
+            try
+            {
+                IRestResponse response = restClient.Execute(restRequest);
+                if (response.IsSuccessful)
+                {
+                    Reporter.ToLog(eLogLevel.DEBUG, "Successfully validated execution id " + message);
```
MeniKadosh1, line 235: it's confusing, API will return 'true' if ExecutionID exist and 'false' if not?
this is ValidationAPI and not IsExist API- so need to return 'true' if validation passed (meaning not duplication) and 'false' if failed (meaning there is duplication)

Ginger/GingerCoreNET/Run/RunsetExecutor.cs
```diff
@@ -393,11 +393,24 @@ public async Task RunRunset(bool doContinueRun = false)
...
+                    else
+                    {
+                        if (mSelectedExecutionLoggerConfiguration.PublishLogToCentralDB == ExecutionLoggerConfiguration.ePublishToCentralDB.Yes)
+                        {
+                            AccountReportApiHandler accountReportApiHandler = new AccountReportApiHandler(WorkSpace.Instance.Solution.LoggerConfigurations.CentralLoggerEndPointUrl);
+                            bool isExists = accountReportApiHandler.ExecutionIdValidation((Guid)RunSetConfig.ExecutionID);
+                            if(isExists)
+                            {
+                                Reporter.ToLog(eLogLevel.ERROR, string.Format("Execution id validation failed due to duplication || Runset Name : {0}, Execution Id : {1}, Please provide unique execution id.", RunSetConfig.Name , RunSetConfig.ExecutionID));
+                                return;
```
MeniKadosh1, line 410: We should not return in case duplication found, we need to show WARN on Log that duplicated ExecutionID was used and that new ExecutionID was created instead and share on Log the new ExecutionID used

MeniKadosh1 (approved):

PR #2034, opened 2021-08-18
fixing self healing issue + code enhancement

ravirk91:
Use runset start time to compare last pom update instead of 5 hours
Reprioritize locator keep enabled by default
Bug 9240:Self-Healing phase 1 Testing and Validation_Run Set
Bug 9239:Self-Healing phase 1 Testing and Validation

### DIFF
--- a/Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
+++ b/Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
@@ -213,6 +213,14 @@ namespace Ginger.AnalyzerLib
 
             finally
             {
+                foreach (var issue in mIssues)
+                {
+                    if (issue.CanAutoFix == AnalyzerItemBase.eCanFix.Yes && issue.FixItHandler != null && mAnalyzerUtils.SelfHealingAutoFixIssue)
+                    {
+                        issue.FixItHandler.Invoke(issue, null);
+                    }
+                }
+
                 BusyInProcess = false;
                 mAnalyzerCompleted = true;
                 SetAnalayzeProceesAsCompleted();
--- a/Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
+++ b/Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
@@ -213,13 +213,18 @@ namespace Ginger.AnalyzerLib
 
             finally
             {
-                foreach (var issue in mIssues)
+                if (mAnalyzerUtils.SelfHealingAutoFixIssue)
                 {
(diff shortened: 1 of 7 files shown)


Ginger/GingerCoreNET/Drivers/CoreDrivers/Web/Selenium/SeleniumDriver.cs
```diff
@@ -7047,11 +7047,17 @@ public void HandleActUIElement(ActUIElement act)
                     case ActUIElement.eElementAction.SetText:
                         try
                         {
-                            ClearText(e);
+                            if (e != null)
```
JinendraGhodke, line 7050: we already have null check for e at start of the function. so if e is null we will not reach this code. why we need this additional check ?

Ginger/GingerCoreNET/AnalyzerLib/AnalyzerUtils.cs
```diff
@@ -109,84 +109,93 @@ public List<string> RunBusinessFlowAnalyzer(BusinessFlow businessFlow, Observabl
...
                     }
+
                 }
-           
-            }
 
-            ReportUnusedVariables(businessFlow, usedVariablesInBF, issuesList);
+                ReportUnusedVariables(businessFlow, usedVariablesInBF, issuesList);
+            }
+            catch (Exception ex)
```
JinendraGhodke, line 194: with this we are supressing that exception. so if there is exception in the first flow we will not analyzer remaining and also the user will not know that there is exception unless debug mode is set. what is the exception ? can he handle the reason for the exception instead of try catch ?

JinendraGhodke (changes requested):

Ginger/Ginger/AnalyzerLib/AnalyzerPage.xaml.cs
```diff
@@ -213,6 +213,14 @@ private async Task Analyze()
 
             finally
             {
+                foreach (var issue in mIssues)
+                {
+                    if (issue.CanAutoFix == AnalyzerItemBase.eCanFix.Yes && issue.FixItHandler != null && mAnalyzerUtils.SelfHealingAutoFixIssue)
+                    {
+                        issue.FixItHandler.Invoke(issue, null);
```
JinendraGhodke, line 220: need to use for loop. And if the issue canAutoFix and FixItHandler is called, should be removed from mIssues. 

Otherwise the issue will be fixed but still shown on the Analyzer UI if running the flow

JinendraGhodke (changes requested):
Handle the only comment

JinendraGhodke (approved):

PR #2038, opened 2021-08-19
Fixed Bug 9226_Saving any Item taking too long

arvindgh:
Saving any Item taking too long to complete, when the solution is not in sync with SCM


JinendraGhodke (commented):
RefreshSourceControlStatus is called from 6 different places. can we add task inside this function so it will be handled for all different use cases ?

arvindgh:
> RefreshSourceControlStatus is called from 6 different places. can we add task inside this function so it will be handled for all different use cases ?

Updated as per requested changes

JinendraGhodke (approved):

PR #2040, opened 2021-08-20
Bugs fixed - 9263, 9274

vyasprateek:


Ginger/GingerCoreCommon/Repository/BusinessFlowLib/FlowControl.cs
```diff
@@ -335,18 +335,23 @@ public Guid GetGuidFromValue(bool doNotUseValueCalculated = false)
...
-                        return Guid.Empty;
+                    {
+                        return Guid.Parse(fcValue);
+                    }
                 }
                 else
-                {
-                    return Guid.Parse(fcValue);
-                }
+                    return Guid.Empty;
```
JinendraGhodke, line 354: why this check is needed ? on Action edit page flow control page, if flow control pointed entity is not found we are displaying the value in the error message. With this change we will display empty guid now. So need to cross check

vyasprateek: Exceptions are being thrown here as fcValue evaluates to null and we're trying to execute extension method ( .Contains) on a null object. It's happening whenever user add a new Flow Control and even after adding error msg to console in Catch block, we're returning Empty GUID only.

PR #2050, opened 2021-08-24
Fixed 9294 Windows POM No Default action generated for Element Type - Image

arvindgh:
Added Image type to windows ElementsTypeList

### DIFF
--- a/Ginger/GingerCoreNET/Run/Platforms/PlatformsInfo/WindowsPlatform.cs
+++ b/Ginger/GingerCoreNET/Run/Platforms/PlatformsInfo/WindowsPlatform.cs
@@ -364,6 +364,7 @@ namespace GingerCore.Platforms.PlatformsInfo
                 mElementsTypeList.Add(eElementType.Dialog);
                 mElementsTypeList.Add(eElementType.MenuItem);
                 mElementsTypeList.Add(eElementType.MenuBar);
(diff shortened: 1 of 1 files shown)


Ginger/GingerCoreNET/Run/Platforms/PlatformsInfo/WindowsPlatform.cs
```diff
@@ -364,6 +364,7 @@ public override List<eElementType> GetPlatformUIElementsType()
                 mElementsTypeList.Add(eElementType.Dialog);
                 mElementsTypeList.Add(eElementType.MenuItem);
                 mElementsTypeList.Add(eElementType.MenuBar);
+                mElementsTypeList.Add(eElementType.Image);
                 //mElementsTypeList.Add(eElementType.Browser);
```
JinendraGhodke, line 368: you need to also handle in GetDefaultElementOperation to set default operation for this element type

JinendraGhodke (changes requested):

Ginger/GingerCoreNET/Run/Platforms/PlatformsInfo/WindowsPlatform.cs
arvindgh: Default operation for this type is IsExist

PR #2071, opened 2021-09-03
Added fix for write exception in excel: added null checking for row a…

rajdeep-mukh:
…nd cell and implemented login to create cell when required


JinendraGhodke:
New PR was created and merged by Haim so closing this one

PR #2074, opened 2021-09-07
Added logic to copy value from customized var to origin var in case o…

rajdeep-mukh:
…f value being set to null in RunSet BusinessFlow and added Unit Test for the same

### DIFF
--- a/Ginger/GingerCoreNET/Run/RunsetExecutor.cs
+++ b/Ginger/GingerCoreNET/Run/RunsetExecutor.cs
@@ -207,7 +207,7 @@ namespace Ginger.Run
                     }
                     if (businessFlowRun.BusinessFlowCustomizedRunVariables != null && businessFlowRun.BusinessFlowCustomizedRunVariables.Count > 0)
                     {
-                        ObservableList<VariableBase> allBfVars = BFCopy.GetBFandActivitiesVariabeles(true); 
+                        ObservableList<VariableBase> allBfVars = BFCopy.GetBFandActivitiesVariabeles(true);
                         Parallel.ForEach(businessFlowRun.BusinessFlowCustomizedRunVariables, customizedVar =>
                         {
                             //This is needed to handle updating the outputvariable mappedoutvalues to new style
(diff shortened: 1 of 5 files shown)


Ginger/GingerCoreNET/Run/RunsetExecutor.cs
```diff
@@ -281,6 +281,11 @@ private void CopyCustomizedVariableConfigurations(VariableBase customizedVar, Va
             RepositoryItemBase.ObjectsDeepCopy(customizedVar, originalVar);//need to replace 'ObjectsDeepCopy' with AutoMapper and to map on it which values should be overiden
             originalVar.DiffrentFromOrigin = customizedVar.DiffrentFromOrigin;
             originalVar.MappedOutputVariable = customizedVar.MappedOutputVariable;
+            //Fix for Bug 9310 - Beeline - Empty variable are not being saved in Run Configuration
```
JinendraGhodke, line 284: we don't add ticket numbers or bugs in the comment. 
You can explain what the particular piece of code is doing if it is needed but not the tickets

Ginger/GingerCoreNETUnitTest/RunTestslib/VariableInputOutoutMappingTest.cs
```diff
@@ -93,10 +93,30 @@ public void OutputVariableOldMappingValueRunsetConfigLoadTest()
...
+            ObservableList<GingerCore.Variables.VariableBase> lstVariables = targetBusinessFlow.Activities[0].Variables;
+            //Act
+            RunSetConfig runSetConfig = (from x in WorkSpace.Instance.SolutionRepository.GetAllRepositoryItems<RunSetConfig>() where x.Name == "Check Variables Run Set" select x).SingleOrDefault();
+            WorkSpace.Instance.RunsetExecutor.RunSetConfig = runSetConfig;
+            WorkSpace.Instance.RunsetExecutor.RunsetExecutionEnvironment = WorkSpace.Instance.SolutionRepository.GetAllRepositoryItems<ProjEnvironment>().FirstOrDefault();
+            WorkSpace.Instance.RunsetExecutor.InitRunners();
+            ObservableList<GingerCore.Variables.VariableBase> lstRunSetVariables = runSetConfig.GingerRunners[0].BusinessFlowsRunList[0].BusinessFlowCustomizedRunVariables;
+            
+            //Assert
+            Assert.AreEqual(null, lstRunSetVariables[0].Value, "Var A is supposed to be equal to null");
```
JinendraGhodke, line 117: we agreed to add 3 variables right ?
1 different from origin and null value
2. Different from origin and non null value
3. Same as origin

for the 2nd use case we are missing the assert

JinendraGhodke (changes requested):

Ginger/GingerCoreNETUnitTest/RunTestslib/VariableInputOutoutMappingTest.cs
rajdeep-mukh: For 2nd case I have added Assert.AreNotEqual on line 118, for 3rd case, I found that variables which are not changed from origin are not stored within RunSet Variables. Hence, thought to skip that.

JinendraGhodke (approved):

PR #2082, opened 2021-09-23
Fixed issue when running webserver action

ranjit5544:


Maheshkale447 (approved):

JinendraGhodke (changes requested):
it was working for user on old Ginger and started failing on 3.9. So we need to understand what is the reason.

ranjit5544:
> it was working for user on old Ginger and started failing on 3.9. So we need to understand what is the reason.

After code is moved to GingerCoreNet this issue occurred. The instance was not getting create. Now, we are creating the instance.

Maheshkale447:
Another PR is opened for this changes

PR #2089, opened 2021-09-28
Feature/self healing phase2

ravirk91:
Feature/self healing phase2

### DIFF
--- a/Ginger/GingerCoreCommon/Repository/BusinessFlowLib/Activity.cs
+++ b/Ginger/GingerCoreCommon/Repository/BusinessFlowLib/Activity.cs
@@ -59,9 +59,9 @@ namespace GingerCore
     }
     public enum eHandlerMappingType
     {
+        None = 0,
         [EnumValueDescription("All Available Error Handlers ")]
-        AllAvailableHandlers = 0,
-        None = 1,
+        AllAvailableHandlers = 1,
         [EnumValueDescription("Specific Error Handlers")]
         SpecificErrorHandlers = 2
     }
--- a/Ginger/GingerCoreCommon/Repository/BusinessFlowLib/ErrorHandler.cs
+++ b/Ginger/GingerCoreCommon/Repository/BusinessFlowLib/ErrorHandler.cs
@@ -16,6 +16,7 @@ limitations under the License.
 */
 #endregion
 
+using Amdocs.Ginger.Common;
 using Amdocs.Ginger.Common.Enums;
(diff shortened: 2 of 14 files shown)


Ginger/GingerCoreCommon/Repository/BusinessFlowLib/ErrorHandler.cs
```diff
@@ -42,6 +66,27 @@ public eHandlerType HandlerType
...
+        public eErrorHandlerPostExecutionAction ErrorHandlerPostExecutionAction
+        {
+            get { return mErrorHandlerPostExecutionAction; }
+            set { if (mErrorHandlerPostExecutionAction != value) { mErrorHandlerPostExecutionAction = value; OnPropertyChanged(nameof(ErrorHandlerPostExecutionAction)); } }
+        }
+
+
+        private eTriggerErrorType mTriggerErrorType;
+        [IsSerializedForLocalRepository]
+        public eTriggerErrorType TriggerErrorType
```
JinendraGhodke, line 81: Let's rename it tp TriggerType and also the enum as eTriggerType.

Ginger/GingerCoreCommon/Repository/BusinessFlowLib/Activity.cs
```diff
@@ -59,9 +59,9 @@ public enum eItemParts
     }
     public enum eHandlerMappingType
     {
+        None = 0,
```
JinendraGhodke, line 62: i think changing the enum default will break existing flows ?

To verify open any existing business flow, and the activity error handler mapping should be set to All available error handler by default

Ginger/GingerCoreNET/Run/GingerRunner.cs
```diff
@@ -1236,6 +1250,8 @@ private void RunActionWithRetryMechanism(Act act, bool checkIfActionAllowedToRun
                 string actionStartTimeStr = string.Empty;
                 while (act.Status != Amdocs.Ginger.CoreNET.Execution.eRunStatus.Passed)
                 {
+                    act = (Act)CurrentBusinessFlow.CurrentActivity.Acts.CurrentItem;
```
JinendraGhodke, line 1253: i think this should be inside the error handler method itsel.

JinendraGhodke (changes requested):
Please check the comments + test if the HTML report is generated correctly for different combinations + add unit tests

PR #2091, opened 2021-09-29
Added Regex to handle tokenizing issues for GetSetDataUsed function

rajdeep-mukh:


uzanh (commented):
need to consolidated this function with writeData(), avoiding duplicated code

PR #2092, opened 2021-09-29
Linux migration/remove generate file from template to legacy

uzanh:


MeniKadosh1 (dismissed):

JinendraGhodke:
For legacy actions usually we have some new action.  
But for actions like this ambar suggested that we should complete hide them from add action panel. The existing actions will still execute but user won't be able to add new one. 

@MeniKadosh1  please discuss with Ambar on this.
May be we can another interface for this type of actions

Ginger/Ginger/AutomatePageLib/AddActionMenu/ActionsFactory.cs
```diff
@@ -181,6 +181,10 @@ static Act GenerateSelectedAction(Act selectedAction, Context mContext)
                             selectedAction.Platform = mContext.Platform;
                         }
                         instance = ((IObsoleteAction)selectedAction).GetNewAction();
+                        if(instance == null)
+                        {
```
MeniKadosh1, line 185: Add check if new Action exist for the Obsolete Action and if not then show new message which not recommending to use the Action and asking if still to add it Yes/No

MeniKadosh1 (approved):

PR #2190, opened 2021-12-16
Updated docker files

arvindgh:
updated docker files with dependencies like git, firefox and chrome for ginger runtime


Ginger/Ginger/DockerFile
```diff
@@ -1,7 +1,7 @@
  
 FROM mcr.microsoft.com/dotnet/framework/runtime:4.7.1 AS runtime
 
-COPY ./Ginger ./Ginger
+COPY ./bin/Debug/ ./Ginger
```
ambarkhan, line 4: change it to bin release

arvindgh: Done

PR #2197, opened 2021-12-23
Feature/map to alm wizard

uzanh:


Ginger/Ginger/ALM/Repository/ZephyrEnt_Repository.cs
```diff
@@ -576,7 +578,36 @@ public override void UpdateBusinessFlow(ref BusinessFlow businessFlow)
         {
             return; // todo refresh
         }
-
-        
+        public override Page GetALMTestSetsTreePage(string importDestinationPath = "")
```
MeniKadosh1, line 581: what happens if we try to do manual mapping for ALM type which is not Zephyr?

MeniKadosh1 (approved):

Ginger/Ginger/ALM/Repository/ZephyrEnt_Repository.cs
uzanh: override method GetALMTestSetsTreePage

PR #2203, opened 2021-12-29
Fixed issue related to memory leak for Excel action

ranjit5544:


Ginger/GingerCoreNET/Run/Excels/ExcelNPOIOperations.cs
```diff
@@ -354,5 +353,13 @@ public List<string> GetSheets(string fileName)
             }
             return sheets.OrderBy(itm => itm).ToList();
         }
+
+        public void Dispose()
+        {
+            mSheet = null;
+            mWorkbook.Close();
+            mWorkbook = null;
+            GC.Collect();
```
JinendraGhodke, line 362: try to comment GC.collect() and see if it will still work. 
If we are disposing the objects  then garbage collector is automatically called after some interval.

JinendraGhodke (changes requested):

PR #2210, opened 2021-12-31
Bug fix/remote execution Account report fixes

ravirk91:
Bug 9427 fixes
Bug 9428 fixes

### DIFF
--- a/Ginger/Ginger/Run/RunSetsExecutionsHistoryPage.xaml.cs
+++ b/Ginger/Ginger/Run/RunSetsExecutionsHistoryPage.xaml.cs
@@ -46,6 +46,7 @@ namespace Ginger.Run
     {
         ObservableList<RunSetReport> mExecutionsHistoryList = new ObservableList<RunSetReport>();
         ExecutionLoggerHelper executionLoggerHelper = new ExecutionLoggerHelper();
+        public bool AutoLoadExecutionData = false;
         public ObservableList<RunSetReport> ExecutionsHistoryList
         {
             get { return mExecutionsHistoryList; }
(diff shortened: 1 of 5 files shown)


Ginger/Ginger/Run/RunSetsExecutionsHistoryPage.xaml.cs
```diff
@@ -104,6 +105,23 @@ private void SetGridView()
             grdExecutionsHistory.AddToolbarTool("@Delete_16x16.png", "Delete Selected Execution Results", new RoutedEventHandler(DeleteSelectedExecutionResults));
             grdExecutionsHistory.AddToolbarTool("@Trash_16x16.png", "Delete All Execution Results", new RoutedEventHandler(DeleteAllSelectedExecutionResults));
             grdExecutionsHistory.RowDoubleClick += OpenExecutionResultsFolder;
+
+            if (mExecutionHistoryLevel == eExecutionHistoryLevel.SpecificRunSet)
```
MeniKadosh1, line 109: why only for "SpecificRunset" page to have the auto update and not also for the general page?

Ginger/Ginger/Run/RunSetsExecutionsHistoryPage.xaml.cs
```diff
@@ -104,6 +105,23 @@ private void SetGridView()
...
+            {
+                grdExecutionsHistory.AddCheckBox("Auto Load Execution History", new RoutedEventHandler(AutoLoadExecutionHistory));
+            }
+        }
+
+        private void AutoLoadExecutionHistory(object sender, RoutedEventArgs e)
+        {
```
MeniKadosh1, line 116: not critical for this case but always better to use binding for connecting between UI control and backend field

Ginger/GingerCoreNET/Run/RunListenerLib/CenteralizedExecutionLogger/AccountReportExecutionLogger.cs
```diff
@@ -53,15 +53,15 @@ public AccountReportExecutionLogger(Context context)
         }
 
         #region RunSet
-        public void RunSetStart(RunSetConfig runsetConfig)
+        public async Task RunSetStart(RunSetConfig runsetConfig)
         {
```
MeniKadosh1, line 57: problem i have here is why this async-await was added only for Runset start/end and not also to rest entities like BusinessFlow/Activity/Action start/end as well? what is the different need here? why it was needed in first place? was this change tested for all cases, from UI and from CLI modes to understand impact?

I will set meeting to discuss it together with Sudarshan.

MeniKadosh1 (changes requested):

MeniKadosh1 (approved):

PR #2233, opened 2022-01-13
Account report variable details

sudarshan0212:


MeniKadosh1 (commented):
what about the updated Contract? needs to be part of Ginger PR, no?
please publish first the nugget then add it to this PR

MeniKadosh1 (approved):

PR #2250, opened 2022-02-01
Fixed 11017 Agent rename issue

arvindgh:


JinendraGhodke (commented):
Arvind, can you cross check if there is possibility to face similar issue in other classes that are refactored ?

JinendraGhodke (commented):
Also the PR needs to be for Beta branch

PR #2321, opened 2022-02-24
Remove core common dependencies

arvindgh:


Maheshkale447 (approved):

Ginger/GingerUtils/OS/OperatingSystemBase.cs
```diff
@@ -75,5 +76,32 @@ private static OperatingSystemBase GetOperatingSystem()
...
+            {
+                if (relativePath.TrimStart().StartsWith("~"))
+                {
+                    string fullPath = relativePath.TrimStart(new char[] { '~', '\\', '/' });
+                    fullPath = Path.Combine(mSolutionFolderPath, fullPath);
+                    return CurrentOperatingSystem.AdjustFilePath(fullPath);
+                }
+            }
+            catch (Exception ex)
+            {
```
JinendraGhodke, line 100: Avoid empty catch, please add logs

Ginger/GingerCoreCommon/Repository/BusinessFlowLib/FlowControl.cs
```diff
@@ -256,7 +256,7 @@ public void CalculateCondition(BusinessFlow BusinessFlow, Environments.ProjEnvir
                 {
                     if (VE.Value.Contains("{Actual}"))
                     {
-                        if ((ARC.Actual != null) && StringManager.IsNumeric(ARC.Actual))
+                        if ((ARC.Actual != null) && ARC.Actual.All(char.IsDigit))
```
JinendraGhodke, line 259: in the old function we were checking char.IsDigit(c) && c != '.'
with the new condition check for != '.' is missing.

Flow control is very sensitive area, can we keep same conditions ?

JinendraGhodke (changes requested):

Ginger/GingerCoreCommon/Repository/BusinessFlowLib/FlowControl.cs
arvindgh: Done

Ginger/GingerUtils/OS/OperatingSystemBase.cs
arvindgh: We do have access to Reporter in this project so removed try catch so that exception will be sent to the caller and can be logged there.

PR #2326, opened 2022-03-01
fix the issue where columns with the '.' or '/' characters in their h…

amitamir:
…eader would not display the cell content of that columns


amitamir:
made more changes, will create a new pull request

PR #2330, opened 2022-03-02
Use act InputValues for serialized feilds on acts

arvindgh:


Maheshkale447 (changes requested):
implement codacy comments

PR #2347, opened 2022-03-23
fixing issues related to Virtual Activity

Maheshkale447:


Ginger/GingerCoreCommon/ReporterLib/UserMsgsPool.cs
```diff
@@ -222,7 +222,7 @@ public static void LoadUserMsgsPool()
...
-            Reporter.UserMsgsPool.Add(eUserMsgKey.AskIfWantsToChangeeRepoItem2, new UserMsg(eUserMsgType.WARN, "Change Repository Item", "The item '{0}' may be used in many places." + Environment.NewLine + Environment.NewLine + "Are you sure you want to {1} it?" + Environment.NewLine + Environment.NewLine + "Note: Anyway the changes won't affect the linked instances of this item", eUserMsgOption.YesNo, eUserMsgSelection.No));
+            Reporter.UserMsgsPool.Add(eUserMsgKey.AskIfWantsToChangeeRepoItem2, new UserMsg(eUserMsgType.WARN, "Change Repository Item", "The item '{0}' may be used in many places." + Environment.NewLine + Environment.NewLine + "Are you sure you want to {1} it?" + Environment.NewLine + Environment.NewLine + "Note: Changes will reflect in linked instances of this item", eUserMsgOption.YesNo, eUserMsgSelection.No));
```
JinendraGhodke, line 225: Rename the message key AskIfWantsToChangeeRepoItem2 to AskIfWantsToChangeLinkedRepoItem

Also the message, let's keep it simple like below

"Modifying linked item will auto update all linked instances."  
 Are you sure you want to {1} it?"

Ginger/Ginger/Activities/ActivityPage.xaml.cs
```diff
@@ -492,15 +494,20 @@
...
 
-        private void SharedRepoSaveBtn_Click(object sender, RoutedEventArgs e)
+        private async void SharedRepoSaveBtn_Click(object sender, RoutedEventArgs e)
         {
             if (mPageViewMode == Ginger.General.eRIPageViewMode.SharedReposiotry)
             {
                 if (SharedRepositoryOperations.CheckIfSureDoingChange(mActivity, "change") == true)
                 {
                     WorkSpace.Instance.SolutionRepository.SaveRepositoryItem(mActivity);
+                    await FindUsages().ConfigureAwait(false);
```
JinendraGhodke, line 504: 1. the method is async so we don't need ConfigureWait(false)
2.  we should call this method only if the mActivity type is linked right ?
3. Rename find usage method to Update Linked instances 

Can you test it with Big solution.  i think saving will just be stuck for long time. althought it will async we need some indication/progress bar for user that we are updating linked items

JinendraGhodke (changes requested):

Ginger/GingerCoreCommon/ReporterLib/UserMsgsPool.cs
Maheshkale447: chaneged enum name and updated the message

Ginger/Ginger/Activities/ActivityPage.xaml.cs
Maheshkale447: 1. the method is async so we don't need ConfigureWait(false),
-  wanted it to run it on background thread so marked it as false, by default its set to true. I removed it and also remoed dispatcher as we don't need it anymore
2. we should call this method only if the mActivity type is linked right ?
- no, as we don't have type flag as link in shared repository, so whenerver we update shared repository activity, it will try to updated all linked instance of matching parent Guid in all Bussiness flow. 
3. Rename find usage method to Update Linked instances
- done

Also tested it with big solution, worked fine. Also added message in status bar that its updating and saving linked instances

PR #2350, opened 2022-03-23
Updated Executer Service Contracts to latest

vyasprateek:


Ginger/GingerRuntime/GingerRuntime.csproj
```diff
@@ -40,7 +40,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Ginger.ExecuterService.Contracts" Version="4.3.1" />
```
MeniKadosh1, line 43: i will approve but why we have reference from GingerRuntime to the Contracts? should be only from GingerCoreNET no?

MeniKadosh1 (approved):

PR #2385, opened 2022-04-14
Ocr Implementation

rajdeep-mukh:


Ginger/GingerCore/GingerOCR/GingerOcrOperations.cs
```diff
@@ -0,0 +1,522 @@
...
+                    }
+                }
+            }
+            return resultTxt;
+        }
+
+        private static string GetResultTextFromByteArray(string label, string resultTxt, byte[] byteArray)
+        {
+            Page pageObj = GetPageObjectFromByteArray(byteArray);
+            using (pageObj)
```
Maheshkale447, line 278: can we create pageObj inside using, if not, lets assign null to in finally block so garbage collector will free its memory

Maheshkale447 (changes requested):

Maheshkale447 (approved):

Maheshkale447 (approved):

PR #2408, opened 2022-05-02
Fixed Ginger crash while sending data to centralized db

arvindgh:
Fixed screenshot move issue + ginger crash while sending data to centralized db using during execution option

### DIFF
--- a/Ginger/GingerCoreNET/Run/RunListenerLib/LiteDBRepository.cs
+++ b/Ginger/GingerCoreNET/Run/RunListenerLib/LiteDBRepository.cs
@@ -95,7 +95,7 @@ namespace Amdocs.Ginger.CoreNET.Run.RunListenerLib
                     {
                         if (File.Exists(completeSSPath))
                         {
-                            File.Delete(completeSSPath);
+                            continue;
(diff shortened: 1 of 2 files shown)


Ginger/GingerCoreNET/Run/RunListenerLib/LiteDBRepository.cs
```diff
@@ -95,7 +95,7 @@ public override object SetReportAction(GingerCore.Actions.Act action, Context co
                     {
                         if (File.Exists(completeSSPath))
                         {
-                            File.Delete(completeSSPath);
+                            continue;
```
Maheshkale447, line 98: why delete was removed, were we deleting the same SS or it was for older SS?

Maheshkale447 (changes requested):

Ginger/GingerCoreNET/Run/RunListenerLib/LiteDBRepository.cs
arvindgh: yes, we were deleting the same SS

PR #2414, opened 2022-05-04
Bug Fixes

rajdeep-mukh:


Ginger/GingerCore/Actions/ActSikuli.cs
```diff
@@ -164,18 +164,16 @@ public string ProcessNameForSikuliOperation
...
 
-        private string mPatternSimilarity = "70";
-
         [IsSerializedForLocalRepository]
         public string PatternSimilarity
         {
             get
             {
-                return mPatternSimilarity;
+                return GetOrCreateInputParam(nameof(PatternSimilarity)).Value;
```
JinendraGhodke, line 172: if storing field on input params then input params are by serialized from Act.cs inputs values list. 
So IsSerializedForLocalRepository is not needed for this field.

Ginger/Ginger/Actions/ActionEditPages/ValidationDBPage.xaml.cs
```diff
@@ -409,6 +409,8 @@ private void SetVisibleControlsForAction()
                 lblInsertJson.Visibility = Visibility.Collapsed;
                 gridInsertJson.Visibility = Visibility.Collapsed;
                 imgHelpSql.Visibility = Visibility.Collapsed;
+                xKeysStackPanel.Visibility = Visibility.Collapsed;
+                SQLUCValueExpression.ValueTextBox.Text = string.Empty;
```
JinendraGhodke, line 413: why are we clearing this value ?

JinendraGhodke (approved):

PR #2431, opened 2022-05-18
Feature/9465 before open the solution on ginger user always notify with popup regarding version

amitamir:


amitamir:
Needs to update warning message to make it more accurate

amitamir:
updated the message which is shown to user

Maheshkale447 (approved):

PR #2452, opened 2022-06-06
Bug9101(get hash code function fixed)

praneethreddy-4:
function of Get Hashed Data Byte String is resolved to get the correct encoded output


Maheshkale447 (changes requested):
please remove extra added 'Ginger-Shortcut.ink' file

PR #2516, opened 2022-07-03
fixed some spelling errors

amitamir:


amitamir:
the spell checker extension is not stable and also does not provide a way to automatically replace all occurrences of the misspelled words.

PR #2580, opened 2022-07-18
process image is now part of the page frame rather than inside the co…

amitamir:
…ntent panel


Maheshkale447 (changes requested):
Please also change loader element on download solution as well

Maheshkale447 (approved):

PR #2656, opened 2022-09-08
Added Screenshot path to output values

TazeenSudan:
Regarding ticket 16174

### DIFF
--- a/Ginger/GingerCoreNET/ActionsLib/General/ActScreenShot.cs
+++ b/Ginger/GingerCoreNET/ActionsLib/General/ActScreenShot.cs
@@ -128,11 +128,10 @@ namespace GingerCore.Actions
 
             }
 
-            String FileName = "";
-            String timeStamp = DateTime.Now.ToString("dd_MM_yyyy_HH_mm_ss_fff");
+            String FileName = this.Description;
+            String timeStamp = DateTime.Now.ToString("ddMMyyyyHHmmss");
 
             FileName += timeStamp;
-            FileName += this.Description;
 
 
             if (ScreenShots.Count == 0)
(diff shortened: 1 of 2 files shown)


Ginger/GingerCoreNET/ActionsLib/General/ActScreenShot.cs
```diff
@@ -166,6 +165,11 @@ public override void PostExecute()
                     }
                 }
             }
+            string filePath = SaveToFileName + "\\" + FileName + ".jpg";
```
digvijayAMDX, line 168: string filePath = SaveToFileName + "\\" + FileName + ".jpg";
use Path.Combile() method instead of string concatenation

Also in the dictionary object user Key as "ScreenshotFilePath" i.e. somewhat relevant to what we are implementing.

digvijayAMDX (changes requested):

Ginger/GingerCoreNET/ActionsLib/General/ActScreenShot.cs
TazeenSudan: Okay, I'll do the needful. 
Also should we provide both the path as well as  the file name, or only the path for now?

Maheshkale447:
Please change target branch to 'master' from 'Releases/Beta'

Ginger/GingerCoreNET/ActionsLib/General/ActScreenShot.cs
```diff
@@ -166,6 +165,11 @@ public override void PostExecute()
...
+            Dictionary<string, object> outFilePath = new Dictionary<string, object>();
+            outFilePath.Add("FileName", filePath);
```
Maheshkale447, line 170: Will it show multiple output values in case of multiple screenshots?

Ginger/GingerCoreNET/ActionsLib/General/ActScreenShot.cs
Maheshkale447: it should be path + filename

Maheshkale447 (changes requested):

Maheshkale447 (approved):

PR #2731, opened 2022-10-18
Add shared activities to shared activity groups

TazeenSudan:


digvijayAMDX (changes requested):
The method GroupActivitiesHandler, is doing almost same thing in both cases if sender is ucGrid or sender is UcListView, try to optimize the code having a separate method and handle the part there and then make a method call there. and based on type handle the logic.

digvijayAMDX (approved):
Approved

digvijayAMDX (approved):

PR #2763, opened 2022-11-22
set context for action

TazeenSudan:
Business and Activity level variables were not visible in Value Expression Editor in SoapUI Wrapper Action while Configuring properties


digvijayAMDX (approved):
Approved fix this includes fix for bug #9567

PR #2805, opened 2022-12-07
executing flow or runset would cause that entity to be marked as modified

amitamir:
PR description - executing flow or runset would cause that entity to be marked as modified

also removed the start dirty tracking from resumeDirtyTracking method in RepositoryItemBase as I am the only one who uses these methods and when we execute Unit Tests, they should not be tracked but because of the logic that was in that method it started tracking the objects and it caused the unit tests to fail

Thank you for your contribution.
Before submitting this PR, please make sure:
- [ ] PR description and commit message should describe the changes done in this PR
- [ ] Verify the PR is point to correct branch i.e. Release or Beta branch if the code fix is for specific release , else point it to master
- [ ] Latest Code from master or specific release branch is merged to your branch
- [ ] No unwanted\commented\junk code is included
- [ ] No new warning upon build solution
- [ ] Existing unit test cases are passed
- [ ] New Unit tests are added for your development
- [ ] Sanity Tests are successfully executed for Existing Functionality
- [ ] After creating pull request there should not be any conflicts
- [ ] Resolve codacy comments
- [ ] Builds and checks are passed before PR is sent for review
- [ ] Resolve code review comments

### DIFF
--- a/Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
+++ b/Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
@@ -1152,11 +1152,6 @@ namespace Amdocs.Ginger.Repository
 
         public void ResumeDirtyTracking()
         {
-            if (DirtyTracking == eDirtyTracking.NotStarted)
-            {
-                StartDirtyTracking();
-                return;
-            }
 
             if (DirtyTracking == eDirtyTracking.Paused)
             {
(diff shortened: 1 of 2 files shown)


Ginger/GingerCoreCommon/Repository/RepositoryItemBase.cs
```diff
@@ -1152,11 +1152,6 @@ public void PauseDirtyTracking()
 
         public void ResumeDirtyTracking()
         {
-            if (DirtyTracking == eDirtyTracking.NotStarted)
-            {
-                StartDirtyTracking();
-                return;
-            }
```
amitamir, line 1159: remove this as if we haven't start tracking an object, I don't want to track it now. this would cause unit tests to fail due to exception was thrown because of that part
checked all references for that method and all of them are by me from my recent changes, probably was an obsolete method from previous Ginger versions

PR #3264, opened 2023-10-04
Improved the null check to capture actual empty null values.

rathimayur:
Issue occurs when user is trying to capture the null value of Free SQL, in same CS expression user wants to verify the non null value as well.
Thank you for your contribution.
Before submitting this PR, please make sure:
- [x] PR description and commit message should describe the changes done in this PR
- [x] Verify the PR is pointing to correct branch i.e. Release or Beta branch if the code fix is for specific release , else point it to master
- [x] Latest Code from master or specific release branch is merged to your branch
- [x] No unwanted\commented\junk code is included
- [ ] No new warning upon build solution
- [ ] Code Summary\Comments are added to my code which explains what my code is doing
- [x] Existing unit test cases are passed
- [ ] New Unit tests are added for your development
- [ ] Sanity Tests are successfully executed for New and Existing Functionality
- [ ] Verify that changes are compatible with all relevant browsers and platforms.
- [ ] After creating pull request there should not be any conflicts
- [ ] Resolve codacy comments
- [ ] Builds and checks are passed before PR is sent for review
- [ ] Resolve code review comments
- [ ] Help Libray document is updated accordingly for Feature changes.


Ginger/GingerCoreNET/Run/GingerExecutionEngine.cs
```diff
@@ -3136,7 +3136,7 @@ public void CalculateActionFinalStatus(Act act)
                     if (ve.Value.Contains("{Actual}"))
                     {
                         //Replace to 
-                        if ((actReturnValue.Actual != null) && Ginger.Utils.StringManager.IsNumeric(actReturnValue.Actual))
+                        if (!string.IsNullOrWhiteSpace(actReturnValue.Actual) && Ginger.Utils.StringManager.IsNumeric(actReturnValue.Actual))
```
Maheshkale447, line 3139: Whitespace will have " " in output value, output values needed in this format?

Maheshkale447: .

Maheshkale447 (changes requested):

Ginger/GingerCoreNET/Run/GingerExecutionEngine.cs
rathimayur: I tested the use cases where only whitespaces will be there in Output values and tried to evaluate them using CS operation, it works fine.
Though we could use IsNullOrEmpty() also, as in current issue we don't have a scenario for IsNullOrWhiteSpace.
Please suggest if we need to change this to IsNullOrEmpty().

Maheshkale447: We are adding quotes for string values, for numeric its not being added, so empty or whitespace should also have quotes, let me know the use case for this

Maheshkale447 (changes requested):

Ginger/GingerCoreNET/Run/GingerExecutionEngine.cs
rathimayur: User wants to see if the SQL Query output returns either null or some value, so when there is value the evaluate using CS works fine, but when the null is returned from DB, it fails.
user uses below CS for evaluating the value.
{CS Exp=({Actual}.Length == 0) || ({Actual}.Length == 2)}

Maheshkale447 (approved):

Maheshkale447 (changes requested):
Wait with merge

Maheshkale447 (approved):

